PR #7, opened 2025-06-20
Add Quantum Kernel Methods

anthonyrtw:
## Theoretical Background
Kernel methods are a class of algorithm which involve the calculation of the inner product of two feature vectors in a higher dimensional space. For quantum kernel methods, we consider a feature map, $U(x_1)$  and calculate the transition probability for every pair of datapoints:
$$\kappa(x_1, x_2) = |\langle s | U^{\dagger}(x_2) U(x_1) | s \rangle|^{2}$$

Generally, kernel is comutative: $\kappa(x_1, x_2) = \kappa(x_2, x_1)$ and $\kappa(x_1, x_1) = 1$. After the kernel matrix is constructed, optimisation is performed classically using a kernel-based algorithm, such as those available in scikit-learn. For identical input datasets, the kernel matrix is positive semi-definite.

## Description
- Introduce `FeatureMap` class which embeds input datapoints in a quantum circuit.
- Introduce `FidelityKernel` class which leverages `torch` to calculate the transition amplitude of every circuit in parallel.
- Introduce `NKernelAlignment`: Negative kernel alignment. A loss function used for pre-training the kernel matrix. See Section on Advanced usage.

For the `FidelityKernel` class we have the following features:
- When the `FidelityKernel` object is called with input datapoints, it returns a scalar value.
- For two input datasets, the entire kernel matrix is explicitly calculated.
- For a single input dataset, the self inner product is returned. Here, we exploit the kernel matrix's symmetry, and only explicitly calculate the upper triangular of the kernel matrix. The diagonal values are set to 1.
- The output kernel matrix is handled to ensure that it matches the same `type` as the input datasets (`numpy.ndarray` or `torch.Tensor`)
- `FidelityKernel` also supports arguments in other QML algorithms in MerLin such as `unbunching` (Default `True`), shots (Default `None`) and `sampling_method` (Default: `multinomial`).
- If `shots is None`, then the exact transition probabilities are returned.
- One may also specify an argument: `force_psd` : For identical input datasets, the kernel matrix is projected onto the closest positive semi-definite matrix.

```python 
from merlin.quantum_kernels import FeatureMap, FidelityKernel 
from sklearn.svm import SVC
import perceval as pcvl 

# Define data encoding parameters
x1, x2 = P("x1"), P("x2")

# Define circuit and feature map object.
circuit = Circuit(2) // PS(x1) // BS() // PS(x2) // BS()

feature_map = FeatureMap(
    circuit=circuit,
    input_parameters="x",
) 
quantum_kernel = FidelityKernel(
    feature_map=feature_map,
    input_state=[2, 0], 
    unbunching=False,
)

# Constuct the training & test kernel matrices
K_train = quantum_kernel(X_train) 
K_test = quantum_kernel(X_test, X_train) 

# Train classically using sklearn
svc = SVC(kernel='precomputed') # Fastest to consider precomputed kernel matrices.
svc.fit(K_train) 
y_pred = svc.predict(K_test)
```

## Advanced Usage
- The feature map be parameterized with respect to a training parameter and the `FidelityKernel` may be pre-trained via quantum kernel alignment. 
- Here, the `FidelityKernel` acts like a torch module:

```python 
from merlin.quantum_kernels import FeatureMap, FidelityKernel 
from merlin.loss import NKernelAlignment
from sklearn.svm import SVC
import perceval as pcvl 

# Define data encoding parameters
x1, x2 = P("x1"), P("x2")

# Define extra training parameter 
theta = P("theta")

# Define circuit and feature map object.
circuit = Circuit(2) // PS(x1) // BS(theta) // PS(x2) // BS(theta)

feature_map = FeatureMap(
    circuit=circuit,
    input_parameters="x",
    trainable_parameters=["theta"]
) 
quantum_kernel = FidelityKernel(
    feature_map=feature_map,
    input_state=[2, 0], 
    unbunching=False,
)

# Define torch optimizer 
optimizer = torch.optim.Adam(quantum_kernel.parameters(), lr=1e-3)

# Use negative kernel target alignment as a loss function
loss_fn = NKernelAlignment()

for epoch in range(10):
    optimizer.zero_grad()

    # Here, one could do pre-training with batches
    X_batch, y_batch = next(dataloader)
    K_train = quantum_kernel(X_batch)
    
    loss = loss_fn(K_train, y_batch)
    loss.backward()
    optimizer.step()

# Then train with SVC over entire training set
K_train = quantum_kernel(X_train).detach().numpy() 
K_test = quantum_kernel(X_test, X_train).detach().numpy() 

svc = SVC(kernel='precomputed')
svc.fit(K_train) 
y_pred = svc.predict(K_test)
```

## Notes 
If `unbunching == True`, then the kernel matrix loses its commutativity $\kappa(x_1, x_2) \neq \kappa(x_2, x_1)$ and the kernel matrix is not symmetric for identical input datasets. The default value of `unbunching` is `True` to maintain continuity with the rest of MerLin. For this reason, `force_psd` is also default `True`.


jsenellart:
superseeded by #37

PR #13, opened 2025-06-24
Add Perceval Parameter Expression handling

anthonyrtw:
## Summary
- Modify `CircuitConverter` to convert circuit components with Perceval parameter expressions, `Expression`.

## Context
- Perceval components may contain parameter expressions, `Expression`. These are objects which represent arithmetic manipulations of the `Parameter` class.
- Currently, when converting circuits with `Expression`, MerLin ignores the Expression and does not raise any error - leading the user to think that their circuit may have been successfully converted into a tensor.

## Proposed Changes
- Add hidden method, _parse_expression to `CircuitConverter`, which evaluates a Perceval `Expression` by parsing its algebraic representation using the ast library and substituting in the values of its parent parameters.

```python 
from perceval import Circuit, PS, P
from merlin import CircuitConverter
import torch
import numpy as np

# Consider a single phase shifter
circuit = Circuit(2) // PS(2 * P("A"))

# Let's print the unitary for phase of pi and a phase of 2pi
phases = np.pi * torch.tensor([[0.5], [1]])

circuit_graph = CircuitConverter(circuit, ["A"])
print(circuit_graph.to_tensor(phases).real)
```
```
tensor([[[-1.,  0.],
         [ 0.,  1.]],

        [[ 1.,  0.],
         [ 0.,  1.]]])
```


anthonyrtw:
The anticipated changes from Perceval for this PR will be available in Perceval 1.0.0.

PR #24, opened 2025-07-11
add CI workflow and more strict tests

jsenellart:


jsenellart:
@mario-valdivia - can you check my configuration of test workflow?

PR #31, opened 2025-08-05
Add Continuous Benchmarking to Merlin

Zafirmk:


CassNot:
Right now, the threshold to block a PR is 20%, I was wondering if it was already too low and if 10% would be preferrable !
I believe users would notice a 20% slow down.

jsenellart:
did you need to use docker? Running directly the test inside the host VMs (that are already clean and fully identical) would be actually faster and as reliable (since docker itself runs on these VMs)

jsenellart:
has been superseded by #41

PR #36, opened 2025-09-11
Superposition for ff

gleboucher:
Added feedforward with or without state injection which needed: 
- Return amplitudes option in Merlin (returns amplitudes instead of probs)
- Amplitudes encoding with a tensor as an input state
- Feedforward block 
- Documentation for the feedfroward


ben9871 (approved):
all good, approved for merge.

PR #54, opened 2025-10-16
Fix kernel no_bunching mode

CassNot:
## Summary

As highligthed by Anthony, there was an issue with the no_bunching = True mode of the FidelityKernel

- the unbunching statistics are not normalized (presumably after removing the bunching stats from the results.)
- in layer.py, we normalize the results after the SLOS graph (line 605 in  layer.py) -> we should do the same for the FidelityKernel

## Context / Related Issues

The no_bunching mode of the FidelityKernel is wrong

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- First, I added a test (based on Anthony's failing case) and edited it a bit:

1. I added a deterministic RNG so PNR/THR comparisons are reproducible and tightened the Perceval post-processing
2. I rebuilt the Perceval reference unitary without mutating shared circuit objects—fresh forward/backward interferometers mirror the FeatureMap outputs and construct a valid Perceval Unitary for the comparison step

- I updated the kernel to pull normalized probabilities straight from SLOS, ensuring the no‑bunching branch matches Perceval’s thresholded distribution before any sampling noise

## How to test / How to run

```
pytest tests/algorithms/test_kernels.py::TestFidelityKernel::test_kernel_no_bunching
```

## Checklist
- [ ] Code formatted (ruff format)
- [ ] Lint passes (ruff)
- [ ] Static typing passes (mypy) if applicable
- [ ] Unit tests added/updated (pytest)
- [ ] Tests pass locally (pytest)
- [ ] Docs build locally if affected (sphinx)
- [ ] I have added a clear PR title and description

### DIFF
--- a/merlin/algorithms/kernels.py
+++ b/merlin/algorithms/kernels.py
@@ -722,7 +722,9 @@ class FidelityKernel(torch.nn.Module):
 
         # Distribution for every evaluated circuit
         _, amplitudes = self._slos_graph.compute(all_circuits, self.input_state)
-        _, probabilities = self._slos_graph.post_pa_inc(amplitudes, all_circuits)
+        _, probabilities = self._slos_graph.compute_probs_from_amplitudes(
+            amplitudes, all_circuits
+        )
         if probabilities.ndim == 1:
             probabilities = probabilities.unsqueeze(0)
         probabilities = probabilities.to(dtype=self.dtype)
@@ -793,7 +795,9 @@ class FidelityKernel(torch.nn.Module):
 
         kernel_unitary = U @ U_adjoint
         _, amplitudes = self._slos_graph.compute(kernel_unitary, self.input_state)
-        _, probabilities = self._slos_graph.post_pa_inc(amplitudes, kernel_unitary)
+        _, probabilities = self._slos_graph.compute_probs_from_amplitudes(
+            amplitudes, kernel_unitary
+        )
         if probabilities.ndim == 1:
             probabilities = probabilities.unsqueeze(0)
         probabilities = probabilities.to(dtype=self.dtype, device=self.device)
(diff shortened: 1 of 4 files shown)


CassNot:
Comment by @ben9871, verify why SLOS is not normalizing (since commit a2d7c90 on slos_torchscriot)

CassNot:
@ben9871 to answer your comments :

Before the commit a2d7c90 , `compute` normalized the probability vector to sum to 1 whenever `no_bunching` was enabled and no output mapping was applied. 
This commit edited this as follows:
- `compute` now returns raw amplitudes and never performs that renormalization step -> only the new `post_pa_inc` helper reintroduces the probability-level normalization logic when it is explicitly invoked. 
- therefore, callers that relied on `compute` to automatically renormalize probabilities in no-bunching mode must now square the amplitudes (or call `post_pa_inc`) to recover the normalized distribution -> that is what I did in this PR

Now, I believe this change was done because we need to return the amplitudes for applications like the PoolingFF, the FF and it is easier to be able to have the amplitudes and generate probabilities from them

merlin/algorithms/kernels.py
```diff
@@ -711,16 +711,18 @@ def forward(
             all_circuits = U_forward[upper_idx[0]] @ U_adjoint[upper_idx[1]]
 
         # Distribution for every evaluated circuit
-        all_probs = self._slos_graph.compute(all_circuits, self.input_state)[1]
+        _, amplitudes = self._slos_graph.compute(all_circuits, self.input_state)
+        _, probabilities = self._slos_graph.post_pa_inc(amplitudes, all_circuits)
+        if probabilities.ndim == 1:
+            probabilities = probabilities.unsqueeze(0)
+        probabilities = probabilities.to(dtype=self.dtype, device=x1.device)
```
jsenellart: why do we need to move the probabilities while we did not need it before?

CassNot:
> Note: post_pa_inc is not a self-explanatory name, we should change it

Name was changed to compute_probs_from_amplitudes

jsenellart (approved):
good to go!

PR #59, opened 2025-10-23
Add K-MNIST and Fashion-MNIST datasets API

v-espitalier:
## Summary
Add K-MNIST and Fashion-MNIST datasets

## Context / Related Issues
Extend MerLin's datasets API with common ML datasets.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## How to test / How to run
The following commands, run from the MerLin directory, download the datasets and display their metadata: :
```
$ python -m datasets.k_mnist
$ python -m datasets.fashion_mnist
```

### DIFF
--- a/merlin/datasets/fashion_mnist.py
+++ b/merlin/datasets/fashion_mnist.py
@@ -1,0 +1,117 @@
+# MIT License
+#
+# Copyright (c) 2025 Quandela
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


merlin/datasets/fashion_mnist.py
```diff
@@ -0,0 +1,117 @@
...
+        year         = {2017},
+    }""",
+    "creators": [
+        "Zalando Research",
+    ],
+    "year": 2017,
+}
+
+
+def get_data_train_huggingface():
```
CassNot: I have the same comments as for KMNIST:
- we should keep the same name of function get_data_train()
- no HuggingFace personal account should be used here, and Pytorch has a torchvision.datasets.FashionMNIST method [here](https://docs.pytorch.org/vision/stable/generated/torchvision.datasets.FashionMNIST.html)

PR #62, opened 2025-10-29
Experiment Detector support in QuantumLayer (layer.py) and FeatureMap (kernels.py)

philippeschoeb:
## Summary
This PR integrates the necessary support in QuantumLayer (algorithms/layer.py) and FeatureMap (algorithms/kernels.py) in order to consider the detectors (perceval.Detector) defined in perceval.Experiment. In order to do that with FeatureMap, I also had to add experiment (perceval.Experiment) as an argument to its __init__().

Other changes are the tests and documentation corresponding to this integration.

## Context / Related Issues
PR for the release-0.2 or MerLin.

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling
- [ ] Breaking change (requires migration notes)

## Proposed changes
- When an Experiment is used to instantiate a QuantumLayer or FeatureMap, its Detectors are now taken into account.
- Conditions for Detectors to apply (raises RunTimeError if these conditions do not meet and a Detector was specified): 
  1. no_bunching=False
  2. measurement_strategy = MeasurementStrategy.PROBABILITIES or = MeasurementStrategy.MODE_EXPECTATIONS
- Here are the steps followed by the QuantumLayer to consider the Detectors:
  1. A list (length = num_modes) of Detectors is created from the Detectors in Experiment. If the Detectors or even the Experiment are not specified, --> Default: perceval.Detector.pnr() for each mode, which is equivalent to not having a Detector.
  2. Checks if the conditions to apply detectors are met (if Detectors are explicitly specified). Raise Error if not. 
  3. Build a DetectorTransform matrix (merlin/measurement/detectors.py)
  4. Retrieve amplitudes from computation process
  5. Convert to probabilities
  6. Apply detector transform (distribution @ detector_transform_matrix)
  7. If apply_sampling, use pcvl_sampler to get distribution
  8. Apply MeasurementStrategy
  9. Return output
- FeatureMap can now be instantiated with a pcvl.Experiment instead of a pcvl.Circuit or a CircuitBuilder.
- Changed the default value of no_bunching to False in QuantumLayer since it is required to consider Detectors.

## How to test / How to run

**Detector support tests are at tests/measurement/test_detectors.py**

1. Define pcvl.Experiment with your circuit:
```
circuit = pcvl.Circuit(4)
# Add components to your circuit

experiment = pcvl.Experiment(circuit)
```

2. Add detectors to your experiment:
[...]

PR #64, opened 2025-10-30
Photon Loss support in QuantumLayer (layer.py) and FeatureMap (kernels.py)

philippeschoeb:
## Summary
This PR integrates the consideration of the NoiseModel in perceval.Experiment given in input to QuantumLayer & FeatureMap (for quantum kernel methods). This allows a layer or feature map to lose photons based on the NoiseModel's brightness and transmittance.

## Context / Related Issues

This PR builds on top of PR #62. So, all changes from #62 are included in this PR.

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- When an Experiment is used to instantiate a QuantumLayer or FeatureMap, its NoiseModel (stored in experiment.noise) is now taken into account to get photon loss probabilities.
- Condition for photon loss to apply (raises RunTimeError if this condition is not met and a NoiseModel was specified): 
  1. measurement_strategy = MeasurementStrategy.PROBABILITIES (default) or = MeasurementStrategy.MODE_EXPECTATIONS
- Here are the steps followed by the QuantumLayer to support photon loss:
  1. A list (length = num_modes) of survival probabilities is created from the brightness and transmittance in pcvl.NoiseModel in pcvl.Experiment. If the NoiseModel or even the Experiment are not specified, --> Default: survival probability=1 for each mode, which is equivalent to not losing photons.
  2. Checks if the condition to apply photon loss is met (if NoiseModel are explicitly specified). Raise Error if not. 
  3. Build a PhotonLossTransform matrix (merlin/measurement/photon_loss.py)
  4. Retrieve new inferred keys + output_size
  5. Retrieve amplitudes from computation process
  6. Convert to probabilities
  7. Apply photon loss transform (distribution @ photon_loss_transform_matrix)
  8. Apply detector transform (if needed)
  9. If apply_sampling, use pcvl_sampler to get distribution
  10. Apply MeasurementStrategy
  11. Return output

## How to test / How to run

**Photon loss support tests are at tests/measurement/test_photon_loss.py**

1. Define pcvl.Experiment with your circuit:
```
circuit = pcvl.Circuit(4)
# Add components to your circuit

experiment = pcvl.Experiment(circuit)
```

2. Add NoiseModel to your experiment:
```
experiment.noise = NoiseModel(brightness=0.9, transmittance=1.0)
```

3. Define your QuantumLayer with this Experiment:
```
layer = QuantumLayer(
    input_size=2
    input_state=[1, 1, 1, 1]
    experiment=experiment
    input_parameters=["px"]
    measurement_strategy=MeasurementStrategy.PROBABILITIES  # Optional
)
``` 

4. Or define a FeatureMap with this Experiment:
```
feature_map = ML.FeatureMap(
    input_size=2,
    experiment=experiment,
    input_parameters=["px"],
)
kernel = ML.FidelityKernel(
    feature_map=feature_map,
    input_state=[1, 1, 1, 1],
)
```

## Documentation
- [x] Docstrings updated

## Checklist
- [x] Lint passes (ruff)
- [x] I have added a clear PR title and description

## Notes
1. This photon loss support feature was supposed to also be integrated for FeedForward module, but that was not done yet.
2. The photon_survival_probs list will always have the same probability for each mode with this current implementation. However, when support for Loss Channels will be integrated, it will be possible to have different photon_survival_probs for each mode.

### DIFF
--- a/tests/measurement/test_photon_loss.py
+++ b/tests/measurement/test_photon_loss.py
@@ -151,7 +151,7 @@ class TestPhotonLossWithQuantumLayer:
     def test_photon_loss_unbunched(self):
         """No-bunching simulations should run (unless a Detector is specified) and keep binary-valued keys after loss."""
         circuit = pcvl.Circuit(3)
-        circuit.add((0, 1), pcvl.BS())
+        circuit.add((0, 1), pcvl.BS(theta=math.pi / 6))
         experiment = pcvl.Experiment(circuit)
         experiment.noise = pcvl.NoiseModel(brightness=0.7)
 


merlin/algorithms/layer.py
```diff
@@ -348,17 +398,23 @@ def _setup_measurement_strategy_from_custom(
         self, measurement_strategy: MeasurementStrategy
     ):
         """Setup output mapping for custom circuit construction."""
-        # Get distribution size
-        dummy_params = self._create_dummy_parameters()
-        if self.input_state is not None and type(self.input_state) is torch.Tensor:
-            keys, distribution = self.computation_process.compute_superposition_state(
-                dummy_params, return_keys=True
+        if self._photon_loss_transform is None:
```
jsenellart: in which circonstances this test will fail since you control initialization/usage flow? is it an information for the user or for the developper?

merlin/measurement/photon_loss.py
```diff
@@ -0,0 +1,348 @@
...
+            A Fock probability vector after photon loss.
+        """
+        if self._is_identity:
+            return distribution
+
+        matrix: torch.Tensor = cast(torch.Tensor, self._matrix)  # type: ignore[attr-defined]
+        if distribution.dtype != matrix.dtype:
+            matrix = matrix.to(distribution.dtype)
+        if distribution.device != matrix.device:
+            matrix = matrix.to(distribution.device)
```
jsenellart: global note here - for a later review, normally nn module do not move/cast  their internal data to what is needed for processing - if they have not been moved before - they simply fail. So my intuition is that we should not do any of this "smart" moving. what do you think?

merlin/algorithms/layer.py
philippeschoeb: This test will never fail with this current implementation. I thought of it more like a safety net upon further development, to ensure the right flow for QuantumLayer. But it is true that right now, it is useless

PR #74, opened 2025-11-05
added missing packages

QuandelaOSS:


CassNot (approved):
Adding all folders to pyproject for successful building on pip. Should be done for next releases.

PR #181, opened 2026-03-18
cleaner docs build for gtm

mario-valdivia:
## Summary
Cleaner way to integrate GTM tag in docs

## Type of change
- [ X] Documentation update

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -17,3 +17,11 @@
   </script>
   <!-- End Google Tag Manager -->
 {% endblock %}
+
+{% block extrabody %}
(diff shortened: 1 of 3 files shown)


CassNot (approved):
- this new code injects the GTM <noscript> inside Sphinx’s extrabody block
- upon doc building, the source code is 

Image

So the `<noscript>` is indeed rendered immediately after `<body>,` therefore
- the `extrabody` block is injected at the top of `<body>`
- this PR preserves the original behavior (which previously enforced this via regex).

I would approve this PR but before, requiring @copilot review to double check this as I am not so sure of myself regarding GTM

PR #202, opened 2026-04-28
PML-281 QCNNClassifier implementation, documentation, testing and performance notebook

philippeschoeb:
## Summary

This PR introduces the QCNNClassifier model class which is a ready to use model for image classification tasks. The docstring is complete, the tests are complete and there is also a notebook to compare the performance of the QCNNClassifier with the previous Density Matrix QCNN implementation on binary MNIST.

## Related Issue
   
Related Jira: PML-281

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- Added the QCNNClassifier in merlin __init__.py so it is easily accessible.
- Small fix in layer.py to usage with MeasurementStrategy.partial()
- Small gradient propagation fix in partial_measurement.py
- Updated pre-commit file to use updated ruff version (like on the actual repo: ruff v0.15.12)

## How to test / How to run

1. Command lines

```
pytest tests/models/test_qcnn.py
```

## Documentation
- [X] User docs updated (Sphinx)
- [X] Examples / notebooks updated
- [X] Docstrings updated
- [X] Updated the API

## Checklist
- [X] PR title includes Jira issue key (e.g., PML-126)
- [X] "Related Jira ticket" section includes the Jira issue key (no URL)
- [X] Code formatted (ruff format)
- [X] Lint passes (ruff)
- [X] Static typing passes (mypy) if applicable
- [X] Unit tests added/updated (pytest)
- [X] Tests pass locally (pytest)
- [X] Tests pass on GPU (pytest)
- [X] Test coverage not decreased significantly
- [X] Docs build locally if affected (sphinx)
- [X] With this command: 

- [X] New public classes/methods/packages are added in the API following the methodology presented in other files.
- [X] Dependencies updated (if needed) and pinned appropriately
- [X] PR description explains what changed and how to validate it


merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+        if stages is not None and type(stages) is not list:
+            raise TypeError("stages must be None or have the list type.")
+
+        self._resolved_stages = self.resolve_stages()
+        self.layers = self.build_qcnn_model()
+
+    @property
+    def resolved_stages(self) -> list[QCNNClassifier._Stage]:
+        """Validated stage sequence used to build the model."""
+        return self._resolved_stages
```
ben9871: This returns the internal mutable list. That makes `resolved_stages` inspectable, but not effectively read-only: a caller can do `model.resolved_stages.append(...)` after construction. 

Since `self.layers` is already built, `summary()` / `export_config()` can then describe a different architecture than the one actually executed.
i.e behaves like such

model = QCNNClassifier((4, 4), 2)
model.resolved_stages.append(QCNNClassifier.QDense())

assert list(model.layers._modules) == ["QConv_1", "QPool_1", "QDense", "Readout"]
assert [type(s).__name__ for s in model.resolved_stages] == [
    "QConv",
    "QPool",
    "QDense",
    "QDense",
]

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+        if self.stages is None or not self.stages:
+            resolved_stages: list[QCNNClassifier._Stage] = []
+            # Default stages
+            resolved_stages.append(QCNNClassifier.QConv(2, 2))
+            resolved_stages.append(QCNNClassifier.QPool(2))
+            resolved_stages.append(QCNNClassifier.QDense())
+            return resolved_stages
+
+        # If stages were specified
+        resolved_stages: list[QCNNClassifier._Stage] = self.stages
```
ben9871: This aliases the user-provided `stages` list. If the caller mutates their list after model construction, the model's `resolved_stages`, `summary()`, and `export_config()` change, but the executable `self.layers` does not. That creates a mismatch between metadata and behavior.

evidence:
stages = [QCNNClassifier.QConv(2, 2), QCNNClassifier.QDense()]
model = QCNNClassifier((4, 4), 2, stages=stages)

stages.append(QCNNClassifier.QPool(2))

assert [type(s).__name__ for s in model.resolved_stages] == [
    "QConv",
    "QDense",
    "QPool",
]
assert list(model.layers._modules) == ["QConv_1", "QDense", "Readout"]

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+            Validated stage sequence used to build the QCNN.
+
+        Raises
+        ------
+        ValueError
+            If the stage order or stage parameters are incompatible with the
+            current register dimensions, or if an unknown stage type is present.
+        """
+        # Check if stages is None or empty
+        if self.stages is None or not self.stages:
```
ben9871: This treats `stages=[]` the same as `stages=None`. I think that weakens the ticket contract: `None` means "use the documented default", but if `stages` is provided, it should fully determine the architecture.

An empty list is provided but does not define a valid QCNN because it has no final `QDense`. Can we make only `stages is None` resolve to the default and make `[]` raise a clear `ValueError`?

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+            resolved_stages.append(QCNNClassifier.QConv(2, 2))
+            resolved_stages.append(QCNNClassifier.QPool(2))
+            resolved_stages.append(QCNNClassifier.QDense())
+            return resolved_stages
+
+        # If stages were specified
+        resolved_stages: list[QCNNClassifier._Stage] = self.stages
+        # Check that only last stage is QDense
+        for i, stage in enumerate(self.stages):
+            if stage.type == QCNNClassifier._QCNNStageTypes.QDense and i != (
```
ben9871, line 332: This reads `stage.type` before validating that `stage` is a supported stage object. For `stages=[object()]`, the constructor raises `AttributeError: 'object' object has no attribute 'type'` instead of the clear `ValueError` promised by the API.

Can we validate `isinstance(stage, (QCNNClassifier.QConv, QCNNClassifier.QPool, QCNNClassifier.QDense))` before reading `stage.type`?

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+        for stage in self._resolved_stages:
+            stage_cfg = {"type": stage.type.value}
+            if isinstance(stage, QCNNClassifier.QConv):
+                stage_cfg["kernel_size"] = stage.kernel_size
+                stage_cfg["stride"] = stage.stride
+            elif isinstance(stage, QCNNClassifier.QPool):
+                stage_cfg["kernel_size"] = stage.kernel_size
+            serializable_stages.append(stage_cfg)
+
+        return {
```
ben9871, line 435: Two issues:

1. `"_resolved_stages"` is a private-looking key even though it is part of the public config contract.
2. `input_shape` is a tuple here, but JSON serialization converts it to a list. The constructor currently rejects lists, so `json.loads(json.dumps(model.export_config()))` cannot be used to reconstruct the architecture.

Can we make the config explicitly JSON-round-trippable? For example, use `"resolved_stages"` and either return `input_shape` as a list plus provide `from_config()`, or make the constructor/config loader normalize list/tuple input shapes.

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+            ``(batch_size, basis_size)``. The basis size is determined by
+            ``sum(input_shape)`` modes and two photons.
+        """
+        batch_size = x.shape[0]
+
+        # Prepare amplitude encoded tensor `state_tensor`
+        empty_tensor = torch.tensor([])
+        state_vector = StateVector(empty_tensor, sum(self.input_shape), 2)
+        basis_size = state_vector.basis_size
+        state_tensor = torch.zeros((batch_size, basis_size), dtype=torch.complex64)
```
ben9871: Can we add an explicit CUDA/device test for QCNNClassifier? If the model and input are moved to GPU, the forward pass should keep all intermediate tensors and the final logits on GPU rather than silently creating CPU tensors inside amplitude_encode(). This might be covered already by statevector and layer tests but no harm

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+            raise ValueError("input_shape must be a tuple of size 2.")
+        if type(input_shape[0]) is not int or type(input_shape[1]) is not int:
+            raise ValueError("input_shape must contain integers.")
+        if input_shape[0] != input_shape[1]:
+            raise ValueError(
+                "input_shape must represent a square (i.e. input_shape[0] == input_shape[1])."
+            )
+        if input_shape[0] <= 0 or input_shape[1] <= 0:
+            raise ValueError("input_shape must contain values superior to 0.")
+        if input_shape[0] > max_input_size:
```
ben9871: why is 28 the limit? also, this doesn't seem changeable by the user. We would probably be more justified documenting the memory usage or simulation cost for growing sizes of the QCNN than doing a hard-limit

tests/models/test_qcnn.py
```diff
@@ -0,0 +1,655 @@
...
+        "input_shape": (4, 4),
+        "num_classes": 3,
+        "_resolved_stages": [
+            {"type": "QConv", "kernel_size": 4, "stride": 1},
+            {"type": "QDense"},
+        ],
+    }
+    assert qcnn_classifier.export_config() == expected_config
+
+    # Round trip test
```
ben9871, line 414: This round-trip is useful, but it only tests the in-memory Python dict. The ticket says `export_config()` should return a serializable config sufficient to reconstruct the architecture. Can we add a JSON round-trip here?

Something like:

```python
loaded = json.loads(json.dumps(qcnn_classifier.export_config()))
restored = QCNNClassifier.from_config(loaded)  # or equivalent helper
assert restored.resolved_stages == qcnn_classifier.resolved_stages
```

This would currently catch the tuple/list `input_shape` issue.

tests/models/test_qcnn.py
```diff
@@ -0,0 +1,655 @@
...
+
+    config_stages = _stages_from_config(config)
+    # Build new QCNNClassifier from the config
+    new_qcnn_classifier = QCNNClassifier(
+        config["input_shape"], config["num_classes"], config_stages
+    )
+
+    assert new_qcnn_classifier.input_shape == qcnn_classifier.input_shape
+    assert new_qcnn_classifier.num_classes == qcnn_classifier.num_classes
+    assert new_qcnn_classifier.resolved_stages == qcnn_classifier.resolved_stages
```
ben9871, line 421: Can we add tests that `resolved_stages` is immutable and that mutating the original user-provided `stages` list after construction does not change model metadata?

This matters because the model builds `self.layers` once. If metadata can mutate afterward, `summary()` and `export_config()` can diverge from the executable architecture.

tests/models/test_qcnn.py
```diff
@@ -0,0 +1,655 @@
...
+        QCNNClassifier(accepted_input_shape, float_num_classes_v1)
+
+    with pytest.raises(TypeError, match="num_classes must have int type"):
+        QCNNClassifier(accepted_input_shape, float_num_classes_v2)
+
+    assert qcnn_classifier.input_shape == accepted_input_shape
+    assert qcnn_classifier.num_classes == accepted_num_classes
+
+
+def test_qcnn_stage_api():
```
ben9871, line 127: `stages=[object()]` should raise a clear `ValueError`, not an `AttributeError`. 
add a test for this

ben9871 (changes requested):
I would request changes before merge on a few contract issues:

- `resolved_stages` is mutable and can diverge from the executable `layers`.
- the user-provided `stages` list is aliased, so external mutation after construction changes model metadata.
- limit of 28 for input_size seems arbitrary
- `stages=[]` currently means "use default"; this is documented in the implementation, but conflicts with the ticket wording that provided stages fully determine the architecture.
- `export_config()` is not robust after JSON serialization.
- Need explicit CUDA/device tests

merlin/models/qcnn.py
philippeschoeb: Agreed, the property will instead return `copy.deepcopy(self._resolved_stages)` so the user cannot modify the internal list

merlin/models/qcnn.py
philippeschoeb: Agreed, the method `resolve_stages` now returns `copy.deepcopy(self.stages)` so any change made by the user to stages, afterward will not affect the model's `resolved_stages`, `summary()` and `export_config()`. I'll also add a test for that.

merlin/models/qcnn.py
philippeschoeb: Yes agreed.

merlin/models/qcnn.py
philippeschoeb: Yes

merlin/models/qcnn.py
philippeschoeb: Yes, will accept list and tuple type for input_shape.

merlin/models/qcnn.py
philippeschoeb: Yes

merlin/models/qcnn.py
philippeschoeb: Okay, I agree that the hard limit is restraining but without it, we cannot ensure that the model won't run out of memory. I agree to removing it and I also agree that we should document memory usage but I think that would be in the scope of another ticket

tests/models/test_qcnn.py
philippeschoeb: Yes, good idea

tests/models/test_qcnn.py
philippeschoeb: Yes

tests/models/test_qcnn.py
philippeschoeb: Yes

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+        self.stages = stages
+
+        # Verify inputs
+        max_input_size = 28  # Set to 28 by default (to reevaluate)
+        if not isinstance(input_shape, tuple):
+            raise TypeError("input_shape must have tuple type")
+        if not isinstance(input_shape[0], int) or not isinstance(input_shape[1], int):
+            raise TypeError("input_shape elements must have int type")
+        if len(input_shape) != 2:
+            raise ValueError("input_shape must be a tuple of size 2.")
```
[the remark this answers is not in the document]
philippeschoeb: Fixed

merlin/models/qcnn.py
```diff
@@ -0,0 +1,1129 @@
...
+                    .to_dense()
+                    .unsqueeze(0)
+                    .repeat(batch_size, 1)
+                )
+                # Scale by pixel value
+                pixel_scaling = x[:, :, i, j]
+                pixel_state_tensor = pixel_scaling * repeated_basic_state_tensor
+                # Add this pixel scaled state vector to the amplitude encoded tensor
+                state_tensor += pixel_state_tensor
+
```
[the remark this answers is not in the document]
philippeschoeb: Not visible in the current codebase but I compared the new amplitude encoding approach to the original method and I obtained the same results which is why i replaced the old method directly.

merlin/models/qcnn.py
ben9871: done

merlin/models/qcnn.py
ben9871: done

merlin/models/qcnn.py
ben9871: done

merlin/models/qcnn.py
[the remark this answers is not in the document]
ben9871: done

ben9871 (approved):
all changes finished as requested

PR #204, opened 2026-05-01
PML-283 photonic memristor

LF-Vigneux:
## Summary
Adding a memristive phase shifter feature to the CircuitBuilder.

## Related Issue
PML-283

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes
- Adding a memristive phase shifter component to the CircuitBuilder
- It updates automatically with the previous outputs
- It is not included in the input

## How to test / How to run
```
pytest tests/builder/test_circuit_builder.py tests/algorithms/test_layer_utils.py tests/algorithms/test_layer.py -q
```
## Documentation
- [x] Docstrings updated
- [x] Updated the API


docs/source/api_reference/api/merlin.algorithms.layer.rst
```diff
@@ -295,6 +295,20 @@ The snippet below prepares a basic quantum layer and returns a :class:`~merlin.c
     assert isinstance(probs,ProbabilityDistribution)
     assert isinstance(probs.tensor,torch.Tensor)
 
+
+Memristive phase-shifter
+-------------------------
+While using the  :class:`~merlin.builder.circuit_builder.CircuitBuilder` to create a :class:`~merlin.algorithms.layer.QuantumLayer`, it is important to set the batch size
```
philippeschoeb: Not clear if this applies every time `CircuitBuilder` is used or whether this only applies when using memristors.

docs/source/api_reference/api/merlin.builder.circuit_builder.rst
```diff
@@ -115,6 +115,21 @@ Switching to a Bell-style interferometer is as simple as:
...
+--------------------------
+
+Adds a memristive phase-shifter.
+
+Arguments:
+
+- ``mode`` (``int``): Circuit mode to target.
+- ``update_rule`` (``callable``): Update rule to change the angle of the phase shifter after each forward pass. The function must take two
+   positional arguments: update_rule(state,output). The update rule must also handle batch inputs and return
+   a tensor of size ```[batch_size]``, just like the state parameter. The output will be the same as the forward.
```
philippeschoeb: The output will be the same as the forward of what? This is not clear to me

merlin/builder/circuit_builder.py
```diff
@@ -406,6 +431,51 @@ def _normalize_angle_scale(
...
+        """Add a memristive phase shifter that will update in regards to the update rule after each forward pass.
+
+        Parameters
+        ----------
+        mode : int
+           Circuit mode to target
+        update_rule : ~collections.abc.Callable
+           Update rule to change the angle of the phase shifter after each forward pass. The function must take two
+           positional arguments: update_rule(state,output). The update rule must also handle batch inputs and return
+           a tensor of size `[batch_size]`, just like the state parameter. The output will be the same as the forward.
```
philippeschoeb: Same comment here: it is unclear this references the forward of what

tests/builder/test_circuit_builder.py
```diff
@@ -521,3 +521,273 @@ def test_entangling_layer_models_on_gpu(model):
     assert logits.shape == (2, 4)
     grads = [p.grad for p in layer.parameters() if p.requires_grad]
     assert any(g is not None and torch.any(g != 0) for g in grads if g is not None)
+
```
philippeschoeb, line 530: I might have missed it, but I think you should add api tests where you send invalid update_rules to the builder to test if the right error or warning is raised. Same with invalid `mode` and `initial_state`.

tests/algorithms/test_layer.py
```diff
@@ -1350,6 +1350,351 @@ def forward(self, x):
 
             assert has_trainable_params, "Model should have trainable parameters"
 
+    def test_memrsistive_update(self):
```
philippeschoeb, line 1358: I would add at least one test of "expected" memristor usage on CUDA to verify it runs well on GPU. Refer to other CUDA tests to see the skip if not cuda.is_available logic.

tests/algorithms/test_layer.py
```diff
@@ -1350,6 +1350,351 @@ def forward(self, x):
...
+        name_to_index = (
+            ql.computation_process.converter.memristive_metadata_name_to_index
+        )
+        metadata = ql.computation_process.converter.memristive_metadata
+        current_state = ql.computation_process.converter.memristive_current_state
+        assert metadata[name_to_index["mem1"]]["update_rule"] == update_rule
+        assert metadata[name_to_index["mem2"]]["update_rule"] == update_rule_exp
+        assert current_state[name_to_index["mem1"]] == torch.Tensor([1.2])
+        assert current_state[name_to_index["mem2"]] == torch.Tensor([0.01])
+
```
philippeschoeb, line 1702: I would add a test on "expected" memristor usage with backpropagation of gradient. Check that not all grad values are 0, all grad values are defined, utilizing optimizer.step() to update parameters changes the parameter values.

philippeschoeb (commented):
Verify high-level review as this PR is quite complex. I leave the low-level review for Benjamin or Cassandre who understand Memristors better than me.

Note that the PML indicated is 284 but the right one is 283.

merlin/builder/circuit_builder.py
```diff
@@ -254,17 +270,24 @@ def add_rotations(
...
+
             elif resolved_role == ParameterRole.INPUT:
                 custom_name = f"px{self._input_counter + 1}"
                 self._input_counter += 1
             elif resolved_role == ParameterRole.TRAINABLE:
                 base_name = f"theta_{self._trainable_counter}_{current_mode}"
                 self._trainable_counter += 1
                 custom_name = self._unique_trainable_name(base_name)
+            elif resolved_role == ParameterRole.MEMRISTOR:
+                custom_name = f"mem{self._input_counter + 1}"
```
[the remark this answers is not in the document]
LF-Vigneux: Will be changed in the next commit!

merlin/algorithms/layer.py
```diff
@@ -303,9 +311,23 @@ def __init__(
...
+            circuit_source.builder.memristor_specs
+            if circuit_source.source_type == "builder"
+            else []
+        )
+        self.memristive_history = [
+            [torch.Tensor([i["initial_state"]])] for i in self._memristive_metadata
+        ]
+        self.memristive_state = [
+            torch.Tensor([i["initial_state"]]) for i in self._memristive_metadata
+        ]
```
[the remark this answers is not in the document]
LF-Vigneux: Will be changed in the next commit

merlin/algorithms/layer.py
```diff
@@ -1426,3 +1497,28 @@ def __str__(self) -> str:
...
+        self._memristive_smaller_last_batch = False
+
+        if len(self.memristive_history) == 0:
+            return
+
+        for i in range(len(self.memristive_history)):
+            self.memristive_state[i] = torch.full(
+                [batch_size], self._memristive_metadata[i]["initial_state"]
+            )
+            self.memristive_history[i] = [self.memristive_state[i]]
```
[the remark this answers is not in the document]
LF-Vigneux: Will be changed in the next commit

merlin/algorithms/layer.py
```diff
@@ -958,19 +1005,32 @@ def forward(
...
+
+        # Phase 7: memristive update
+        if len(self.memristive_state) > 0:
+            self.memristive_state = compute_new_memristive_ps_angles(
+                memristive_metadata=self._memristive_metadata,
+                memristive_state=self.memristive_state,
+                output=output,
             )
+            for i in range(len(self.memristive_history)):
+                self.memristive_history[i].append(self.memristive_state[i])
```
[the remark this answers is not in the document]
LF-Vigneux: Here I just deepcopied the output before calling the function since the output needs to be returned as is, not detached. Will be updated in the next commit.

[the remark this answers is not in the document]
LF-Vigneux: After reflection, I detached the tensors as well, since deepcopy wont work by itself. This made me add, with copilot (to be faster) a detach method for PartialMeasurement

merlin/algorithms/layer_utils.py
```diff
@@ -747,6 +749,34 @@ def apply_angle_encoding(
...
+    output: torch.Tensor | PartialMeasurement | merlin.core.state_vector.StateVector | ProbabilityDistribution,
+        The output of the quantum layers
+
+    Returns
+    -------
+    list[torch.Tensor]
+        The new states of all memristive phase shifters
+    """
+    new_memristive_states = []
+    for metadata, state in zip(memristive_metadata, memristive_state, strict=False):
```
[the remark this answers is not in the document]
LF-Vigneux: Done in the next commit!

merlin/pcvl_pytorch/locirc_to_tensor.py
```diff
@@ -392,6 +413,9 @@ def to_tensor(
             )
 
         self.torch_params = input_params
+        self.memristive_current_state = (
+            [] if memristive_current_state is None else memristive_current_state
+        )
```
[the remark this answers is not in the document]
LF-Vigneux: Done in next commit

merlin/core/components.py
```diff
@@ -36,6 +36,7 @@ class ParameterRole(Enum):
     FIXED = "fixed"  # Fixed value, not a parameter
     INPUT = "input"  # Input parameter (fed from data)
     TRAINABLE = "trainable"  # Trainable parameter (nn.Parameter)
+    MEMRISTOR = "memristor"  # Memristor parameters to be updtaed with respect to a rule
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

tests/builder/test_circuit_builder.py
```diff
@@ -521,3 +521,273 @@ def test_entangling_layer_models_on_gpu(model):
...
+        return state + (target - state) / tau
+
+    def sum_outputs(state: torch.Tensor, output: torch.Tensor) -> torch.Tensor:
+        return state + output[:, 2]
+
+    builder.add_angle_encoding()
+    builder.add_rotations(trainable=True)
+    builder.add_rotations(trainable=False)
+    builder.add_superpositions
+
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

tests/builder/test_circuit_builder.py
```diff
@@ -521,3 +521,273 @@ def test_entangling_layer_models_on_gpu(model):
...
+    custom_names = [f"mem{i + 1}" for i in range(4)]
+    num_memristor = 0
+    num_other = 0
+    for component in builder.circuit.components:
+        if component.role is ParameterRole.MEMRISTOR:
+            component.custom_name in custom_names
+            assert isinstance(component, Rotation)
+            num_memristor += 1
+        else:
+            component.custom_name not in custom_names
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

docs/source/api_reference/api/merlin.builder.circuit_builder.rst
```diff
@@ -115,6 +115,21 @@ Switching to a Bell-style interferometer is as simple as:
...
+- ``initial_state`` (``float``): The initial value of the phase shifter. This will be the value used after each :meth:`~merlin.algorithms.layer.QuantumLayer.reset` call.
+- ``name`` (``str``): Prefix used for the generated memristive phase shifter parameter. Defaults to ``"mem"``.
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

docs/source/api_reference/api/merlin.algorithms.layer.rst
```diff
@@ -295,6 +295,20 @@ The snippet below prepares a basic quantum layer and returns a :class:`~merlin.c
...
+of the memristors. In order to do so, the :meth:`~merlin.algorithms.layer.QuantumLayer.reset` method needs to be called. It resets the memristors to their initial state while clearing the history.
+It also prepares the memristors to be ran in a ``batch_size`` (the parameter of the function, defaults to 1) batch size.
+
+The :class:`~merlin.algorithms.layer.QuantumLayer` needs to have a ``batch_size`` input on all forward pass unless :meth:`~merlin.algorithms.layer.QuantumLayer.reset` is called. However if only one batch
+is smaller than the others it can be ran once. The first " smaller batch size" memresitor states will be used. To make another forward call after that, the :meth:`~merlin.algorithms.layer.QuantumLayer.reset` method
+needs to be called again.
```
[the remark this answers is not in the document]
LF-Vigneux: Corrected in next review

docs/source/api_reference/api/merlin.algorithms.layer.rst
LF-Vigneux: Specified in the last commit!

docs/source/api_reference/api/merlin.builder.circuit_builder.rst
LF-Vigneux: Specified in the last commit, In think its more clear now

merlin/builder/circuit_builder.py
LF-Vigneux: Specified in the last commit

tests/algorithms/test_layer.py
LF-Vigneux: Added in the last commit, also verified that the output is still attached to the gradient computation

tests/algorithms/test_layer.py
LF-Vigneux: Done in the last commit

tests/builder/test_circuit_builder.py
LF-Vigneux: Done in the last commit

merlin/builder/circuit_builder.py
```diff
@@ -96,7 +102,9 @@ def __init__(self, n_modes: int):
         self._angle_encoding_specs: dict[str, list[tuple[int, ...]]] = {}
         self._angle_encoding_scales: dict[str, dict[int, float]] = {}
         self._angle_encoding_counts: dict[str, int] = {}
-
+        self.memristor_specs: list[dict] = []
```
ben9871: small thing. The ticket says memristive_specs, not memristor_specs. This is semantically cleaner as you could have multiple memristors. Not a huge thing so I could be convinced otherwise

merlin/algorithms/layer.py
```diff
@@ -1109,6 +1196,18 @@ def to(self, *args, **kwargs):
             if self._detector_transform is not None:
                 self._detector_transform = self._detector_transform.to(device)
 
+            # memristor state and history
+            for state in range(len(self.memristive_history)):
+                for t in range(len(state)):
```
ben9871:
```python
for t in range(len(state)):
```

The loop above iterates `state` as an integer index from `range(...)`, then calls `len(state)`. For a memristive layer, even `layer.to(torch.device("cpu"))` raises:

```text
TypeError: object of type 'int' has no len()
```

This affects the expected device/dtype movement behavior and undermines the CUDA/device acceptance path. Didn't run a CUDA test but this also fails on CPU for me.

merlin/builder/circuit_builder.py
```diff
@@ -406,6 +435,192 @@ def _normalize_angle_scale(
         factor = float(scale)
         return dict.fromkeys(feature_indices, factor)
 
+    @staticmethod
+    def _validate_memristive_update_rule(update_rule: Callable) -> None:
```
ben9871, line 438: The ticket requires `update_rule` to be a Python callable with the logical signature `update_rule(state, output)`. The implementation rejects unannotated callables:

```text
TypeError update_rule first parameter 'state' requires type annotation: state: torch.Tensor
```

This is stricter than the spec and will reject common user patterns such as small local functions or lambdas. For example, while exploring a sequence model, a user may reasonably write:

```python
def exponential_decay(state, output):
    tau = 5.0
    target = output[:, 2]
    return state + (target - state) / tau

builder.add_memristive_ps(
    mode=2,
    update_rule=exponential_decay,
    initial_state=0.0,
)
```

Or, for a quick experiment:

```python
builder.add_memristive_ps(
    mode=2,
    update_rule=lambda state, output: state + 0.1 * output[:, 2],
    initial_state=0.0,
)
```

Both are natural Python APIs for an exploratory ML feature, but the current validator rejects them because it requires inspectable type annotations.

merlin/algorithms/layer_utils.py
```diff
@@ -747,6 +749,34 @@ def apply_angle_encoding(
...
+        The output of the quantum layers
+
+    Returns
+    -------
+    list[torch.Tensor]
+        The new states of all memristive phase shifters
+    """
+    new_memristive_states = []
+    for metadata, state in zip(memristive_metadata, memristive_state, strict=True):
+        new_memristive_states.append(metadata["update_rule"](state, output))
```
ben9871: The ticket says the memristive state shape is `[batch_size]` and the update rule must return a tensor of size `[batch_size]`. The implementation documents that requirement but does not enforce it.

I confirmed that an update rule returning shape `(batch_size, 1)` is accepted and stored as the next state:

```text
bad_shape_after_forward: (2, 1)
```

That can later produce hard-to-debug shape behavior inside the circuit converter. A clear `RuntimeError`/`ValueError` here would make the feature much safer.

ben9871 (changes requested):
Some issues to resolve in API, semantics and additional tests are needed

merlin/builder/circuit_builder.py
LF-Vigneux: Will be changed in next commit

merlin/algorithms/layer.py
LF-Vigneux: Changed in next commit

merlin/algorithms/layer_utils.py
LF-Vigneux: Changed in next commit

docs/source/api_reference/api/merlin.algorithms.layer.rst
```diff
@@ -295,6 +295,20 @@ The snippet below prepares a basic quantum layer and returns a :class:`~merlin.c
     assert isinstance(probs,ProbabilityDistribution)
     assert isinstance(probs.tensor,torch.Tensor)
 
+
+Memristive phase-shifter
+-------------------------
+When using the :class:`~merlin.builder.circuit_builder.CircuitBuilder` that contains a memristive phase-shifter (added with the :meth:`~merlin.builder.circuit_builder.CircuitBuilder.add_memristive_ps` method) to create a :class:`~merlin.algorithms.layer.QuantumLayer`, it is important to set the batch size
+of the memristors. To do so, call the :meth:`~merlin.algorithms.layer.QuantumLayer.reset` method. This resets the memristors to their initial state while clearing their history.
```
CassNot: A bit ambiguous. Is it singular or plural ?

docs/source/api_reference/api/merlin.algorithms.layer.rst
```diff
@@ -295,6 +295,20 @@ The snippet below prepares a basic quantum layer and returns a :class:`~merlin.c
...
+It also prepares the memristors to be run with the specified ``batch_size`` parameter of the function, which defaults to 1.
+
+The :class:`~merlin.algorithms.layer.QuantumLayer` expects all forward passes to use the configured ``batch_size`` unless :meth:`~merlin.algorithms.layer.QuantumLayer.reset` is called again. However, if only one batch
+is smaller than the others, it can be run once. In that case, the first ``smaller batch size`` memristor states will be used. Before making another forward call after that, the :meth:`~merlin.algorithms.layer.QuantumLayer.reset` method
```
CassNot: For clarity, I would add that this is often the case with the last batch of the loader

docs/source/api_reference/api/merlin.builder.circuit_builder.rst
```diff
@@ -115,6 +115,21 @@ Switching to a Bell-style interferometer is as simple as:
...
+--------------------------
+
+Adds a memristive phase-shifter.
+
+Arguments:
+
+- ``mode`` (``int``): Circuit mode to target.
+- ``update_rule`` (``callable``): Update rule to change the angle of the phase shifter after each forward pass of associated the :class:`~merlin.algorithms.layer.QuantumLayer` object. The function must take two
+   positional arguments: update_rule(state,output). The update rule must also handle batch inputs and return
+   a tensor of size ``[batch_size]``, just like the state parameter. The output will be the same as the corresponding :class:`~merlin.algorithms.layer.QuantumLayer` forward.
```
CassNot: I think it would be a good place to describe the expected type and all the rules (number of inputs, types of inputs and outputs) of this method for it to be super clear and avoid errors

docs/source/api_reference/api/merlin.algorithms.layer.rst
```diff
@@ -295,6 +295,20 @@ The snippet below prepares a basic quantum layer and returns a :class:`~merlin.c
     assert isinstance(probs,ProbabilityDistribution)
     assert isinstance(probs.tensor,torch.Tensor)
 
+
+Memristive phase-shifter
+-------------------------
+When using the :class:`~merlin.builder.circuit_builder.CircuitBuilder` that contains a memristive phase-shifter (added with the :meth:`~merlin.builder.circuit_builder.CircuitBuilder.add_memristive_ps` method) to create a :class:`~merlin.algorithms.layer.QuantumLayer`, it is important to set the batch size
```
CassNot: It is not super clear what the reset method does
I understand here that we need to use it if the batch_size changes (but it is more of a usage than a description)
The docstring of the method is super clear but I would add it again here
You can also add a very short block code for a quick demo

docs/source/api_reference/api/merlin.builder.circuit_builder.rst
```diff
@@ -115,6 +115,21 @@ Switching to a Bell-style interferometer is as simple as:
    :width: 200px
    :align: center
 
+**add_memristive_ps**
```
CassNot: I think the description should come last (after `add_superpositions`)

merlin/builder/circuit_builder.py
```diff
@@ -406,6 +434,129 @@ def _normalize_angle_scale(
...
+                f"Expected: update_rule(state: torch.Tensor, "
+                f"output: torch.Tensor | StateVector | ProbabilityDistribution | PartialMeasurement) "
+                f"-> torch.Tensor"
+            )
+
+    def add_memristive_ps(
+        self,
+        mode: int,
+        update_rule: Callable,
+        initial_state: float,
```
CassNot, line 477: Should there be a default initial_state ? For instance pi
I think this is a design question for @ben9871

merlin/builder/circuit_builder.py
```diff
@@ -406,6 +434,129 @@ def _normalize_angle_scale(
...
+                f"update_rule must accept exactly 2 positional arguments, got {len(positional_params)}. "
+                f"Expected: update_rule(state: torch.Tensor, "
+                f"output: torch.Tensor | StateVector | ProbabilityDistribution | PartialMeasurement) "
+                f"-> torch.Tensor"
+            )
+
+    def add_memristive_ps(
+        self,
+        mode: int,
+        update_rule: Callable,
```
CassNot, line 476: Should there be a default update_rule as well (basic exponential_decay for instance) @ben9871 

```
def exponential_decay(state: torch.Tensor, output: torch.Tensor) -> torch.Tensor:
    tau = 5.0
    target = output[:, 2]
    return state + (target - state) / tau
```

The intuition behind is that we do not expect users of builder to come with an update_rule on the first go

@LF-Vigneux , what is your opinion on that ?

CassNot (changes requested):
- the doc builds correctly
- the code is clean
- I was able to test it in the attached notebook to reproduce the memristor introduced in https://github.com/merlinquantum/reproduced_papers/tree/main/papers/qrc_memristor (@LF-Vigneux , you can use it for an additional resource - without the bugs)
- I tried to break it and I almost couldn't so very nice job :) copilot helped me to better break it and here are my takeaways: 

I think some slight fixes needed and some tests that are missing:
BUGS:
- calling `layer.to(torch.float64)` or `layer.to(dtype=torch.float64)` will not update `memristive_state` or `memristive_history` tensor **dtypes** (this should fail now as per the notebook I am attaching - BUG 1)
-  there is one senario I have not ran yet and it is a QuantumLayer that **moves on GPU** and I think it is super important to add one to ensure that all history, states and update_rule are operated on the GPU (**that could be another ticket/PR**) 
- With **multiple memristors**, if an update rule for memristor index 1+ returns a tensor of the **wrong size** (while index 0 is correct), Phase 2 won't catch it. It will be caught in Phase 7 with a less informative error message. (BUG 2 in notebook found by Claude on Copilot)
- `state_dict()` / `load_state_dict()` drops memristive state. When the model is  **saved/loaded**, it will reset the memristive sequence silently. I think it is unexpected behavior for anyone checkpointing a sequence model mid-episode. (it is bug 3 in the notebook I am attaching)


Missing tests:

- check that `reset()` called after `.to(device/dtype)` produces tensors on the correct device
- test that 2 memristive PS with same names in the same layer should produce distinct parameter names
- `memristive_state` and `memristive_history` should be saved by `state_dict()`
- Memristive updates are side effects so calling `model.eval()` should not suppress them
- Verifies that the RuntimeError on batch size mismatch is raised on the very first forward
- Suggested by Claude/Copilot: In a recurrent system the *full* gradient of the loss with respect to a trainable parameter should include **every path** through which the parameter affects the output, otherwise the optimizer could be working with an incomplete gradient signal. The optimiser descends along the direct gradient at each step. If the global minimum requires co-adapting parameters across steps (which could be the point of a recurrent/memristive system), gradient descent may stall far from that minimum. I had not thought about that before but it could be a good point. This is Bug-4 in my notebook. What do we think of that @LF-Vigneux and @ben9871 ? (it could be a separate work as well)

[memristor.ipynb](https://github.com/user-attachments/files/27537556/memristor.ipynb)

LF-Vigneux:
For this point:

there is one senario I have not ran yet and it is a QuantumLayer that moves on GPU and I think it is super important to add one to ensure that all history, states and update_rule are operated on the GPU (that could be another ticket/PR)

The test test_memristor_works_on_cuda on test_layer.py and it should test every basic usage

merlin/builder/circuit_builder.py
LF-Vigneux: I would not put any default on this argument since its the goal of the memristor to define an update function. Also, when writing code, and we see the only the definition of the function :update_rule=exponential_decay. It is not clear exaclty what is done in my opinion

merlin/builder/circuit_builder.py
ben9871: no. I think it might be a bit misleading when they check the value and see how it updated. I think this one should be explicit

merlin/builder/circuit_builder.py
ben9871: I agree with Louis-Felix here. The memristor has a clear goal and is used for specific experiments. Making a default update rule or initial state turns it into a blackbox with low observability and encourages bad user habits imo. It is not something we want to place in every experiment, it must be by design

ben9871:
> * the doc builds correctly
> * the code is clean
> * I was able to test it in the attached notebook to reproduce the memristor introduced in https://github.com/merlinquantum/reproduced_papers/tree/main/papers/qrc_memristor (@LF-Vigneux , you can use it for an additional resource - without the bugs)
> * I tried to break it and I almost couldn't so very nice job :) copilot helped me to better break it and here are my takeaways:
> 
> I think some slight fixes needed and some tests that are missing: BUGS:
> 
> * calling `layer.to(torch.float64)` or `layer.to(dtype=torch.float64)` will not update `memristive_state` or `memristive_history` tensor **dtypes** (this should fail now as per the notebook I am attaching - BUG 1)
> * there is one senario I have not ran yet and it is a QuantumLayer that **moves on GPU** and I think it is super important to add one to ensure that all history, states and update_rule are operated on the GPU (**that could be another ticket/PR**)
> * With **multiple memristors**, if an update rule for memristor index 1+ returns a tensor of the **wrong size** (while index 0 is correct), Phase 2 won't catch it. It will be caught in Phase 7 with a less informative error message. (BUG 2 in notebook found by Claude on Copilot)
> * `state_dict()` / `load_state_dict()` drops memristive state. When the model is  **saved/loaded**, it will reset the memristive sequence silently. I think it is unexpected behavior for anyone checkpointing a sequence model mid-episode. (it is bug 3 in the notebook I am attaching)
> 
> Missing tests:
> 
> * check that `reset()` called after `.to(device/dtype)` produces tensors on the correct device
> * test that 2 memristive PS with same names in the same layer should produce distinct parameter names
> * `memristive_state` and `memristive_history` should be saved by `state_dict()`
> * Memristive updates are side effects so calling `model.eval()` should not suppress them
> * Verifies that the RuntimeError on batch size mismatch is raised on the very first forward
> * Suggested by Claude/Copilot: In a recurrent system the _full_ gradient of the loss with respect to a trainable parameter should include **every path** through which the parameter affects the output, otherwise the optimizer could be working with an incomplete gradient signal. The optimiser descends along the direct gradient at each step. If the global minimum requires co-adapting parameters across steps (which could be the point of a recurrent/memristive system), gradient descent may stall far from that minimum. I had not thought about that before but it could be a good point. This is Bug-4 in my notebook. What do we think of that @LF-Vigneux and @ben9871 ? (it could be a separate work as well)
> 
> [memristor.ipynb](https://github.com/user-attachments/files/27537556/memristor.ipynb)

I think this last point about gradients may be outside the scope of the initial release of this feature. taking the gradient and backpropagating through each timestep will be intensive and we could envision that the user controls how many timesteps they want to backpropagate through. But this changes the nature of the memristor entirely(here it is a kind of reservoir). For me this would be an option of 'memristor_gradient_steps'(placeholder name) which would determine how many timesteps the gradient propagates through(currently 0 which isn't the full gradient but uses less memory). I don't know how difficult it is to do this, if it's easy we can fit it in @CassNot

ben9871 (approved):
Changes I requested are here, however, full gradient backprop through N steps of memristor not implemented. It is out of scope for this ticket in my opinion so approved on my end

CassNot (approved):
Reviewed and approved
The optimisation (and visualisation) will be done in 2 other tickets to separate the problem

merlin/builder/circuit_builder.py
CassNot: Ok, agreed !

PR #205, opened 2026-05-05
PML-284 noisy slos execution plumbing in quantum layer

LF-Vigneux:
## Summary
Plumbing in the layer constructor to propagated the noise to the corresponding variables to be used in the noisy simulation.
The previous API still works but with a new noise model optional parameter in the QuantumLayer

## Related Issue
PML-284

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes
- Added a NoiseGroups class to classify the different noise sources
   - This class will be used in the noisy SLOS
- Added guiderails for unsupported noisy configurations
- The normalized,classified and verified noise is now an attribute of the layer

## How to test / How to run
```
pytest -q
```

## Documentation
- [x] Docstrings updated
- [ ] Updated the API


docs/source/notebooks/QORC-timeseries.ipynb
ben9871, line 1: changes to notebooks are not part of this PR, this is just ruff?

merlin/algorithms/layer_utils.py
```diff
@@ -671,12 +697,51 @@ def setup_noise_and_detectors(
             "Compute amplitudes without detectors and apply a Partial DetectorTransform manually if needed."
         )
```
ben9871, line 701: PML-284 says `brightness` and `transmittance` are post-measurement noise and should remain in the existing photon-loss path via `resolve_photon_loss()`. That works when noise comes from `experiment.noise`, but not when it comes from the new direct `noise_model=` API.

The issue is that `resolve_circuit()` creates a new `Experiment(circuit)` for builder/circuit sources but does not attach the resolved `noise_model` to that experiment. Later, `setup_noise_and_detectors()` calls:

photon_survival_probs, empty_noise_model = resolve_photon_loss(
    experiment, circuit.m
)

```python
layer = ml.QuantumLayer(
    n_photons=2,
    input_size=4,
    builder=circ,
    noise_model=pcvl.NoiseModel(brightness=0.3),
)

print(layer._photon_survival_probs)
print(getattr(layer.experiment, "noise", None))
```

Observed:

```text
[1.0, 1.0, 1.0, 1.0]
None
```

Expected:

```text
[0.3, 0.3, 0.3, 0.3]
NoiseModel(brightness=0.3) available to the photon-loss resolver, or equivalent direct plumbing.
```

merlin/algorithms/layer.py
ben9871, line 423: The ticket explicitly asks for `noise_groups` to propagate through initialization so `ComputationProcessFactory.create()` can route to the correct process subclass later.

The PR adds `noise_groups` to `InitializationContext` and to `ComputationProcessFactory.create()`, but the actual `QuantumLayer` call site never passes it. So it never reads these

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +950,123 @@ def normalize_output_key(
...
+
+    Returns
+    -------
+    NoiseGroups | None
+        The NoiseGroups object that contains the noise sources per category.
+    """
+    # Source noise
+    if noise_model is None:
+        return None
+    source = {
```
ben9871: The ticket says default/`None` values should be excluded because they represent "no noise on this axis." 

Perceval's `NoiseModel` does not expose all defaults as `None`: for example an empty model reports values like `brightness=1`, `indistinguishability=1`, `g2=0`, `g2_distinguishable=True`, `transmittance=1`, `phase_imprecision=0`, and `phase_error=0`.

```python
classify_noise_model(pcvl.NoiseModel())
```

Observed:

```text
NoiseGroups(
  source={"indistinguishability": 1, "g2_distinguishable": True, "g2": 0},
  circuit={"phase_imprecision": 0, "phase_error": 0},
  post_measurement={"transmittance": 1, "brightness": 1},
)
```

This is not the intended classification. It also means `NoiseModel(brightness=0.3)` incorrectly has source and circuit groups, even though only post-measurement noise was requested.

merlin/algorithms/layer_utils.py
```diff
@@ -671,12 +697,51 @@ def setup_noise_and_detectors(
...
 
+    validate_noisy_measurement_strategy(
+        backend,
+        output=measurement_kind.name.lower(),
+        noise_model=noise_model,
+        empty_detectors=empty_detectors,
+    )
+    noise_groups = None if noise_model is None else classify_noise_model(noise_model)
+    # Not implemented errors
+    if noise_groups is not None:
```
ben9871, line 719: PML-284 asks for a clear construction-time `NotImplementedError`, after classification and validation, that identifies the noise groups found and states that the differentiable noisy execution path is not implemented.

Current code raises field-specific messages:

- `merlin/algorithms/layer_utils.py:708`

```python
if noise_groups is not None:
    if noise_groups.source is not None:
        ...
        raise NotImplementedError("The g2 error is not implemented yet")
```

This does not list the classified groups and it only reports the first unsupported field encountered. It also depends on the current misclassification of identity defaults.

merlin/algorithms/layer.py
```diff
@@ -301,6 +315,7 @@ def __init__(
             measurement_strategy=measurement_strategy,
             warnings=noise_and_detectors.detector_warnings,
             return_object=return_object,
```
ben9871, line 318: The ticket marks `return_object=True` with a noise model as out of scope. The current validation does not reject it. For a direct post-measurement-only noise model such as `NoiseModel(brightness=0.3)`, construction succeeds.

ben9871 (changes requested):
The new `tests/algorithms/test_noise_contract.py` file is useful, but it mostly tests all noise fields together or source/circuit NotImplemented cases. These are cases which do not behave as expected in the current code

Missing tests worth adding:

- direct `noise_model=NoiseModel(brightness=0.3)` applies photon survival probabilities, not `[1.0, ...]`
- direct `noise_model=NoiseModel(transmittance=0.4)` applies photon survival probabilities
- `classify_noise_model(pcvl.NoiseModel())` returns no active groups
- `classify_noise_model(pcvl.NoiseModel(brightness=0.3))` contains only post-measurement noise
- `QuantumLayer(..., noise_model=NoiseModel(brightness=0.3)).computation_process.noise_groups` receives the groups after setup
- NotImplemented message lists the group summary
- experiment and direct noise entry points use the same error type/message for unsupported measurement strategies
- `return_object=True` with noise model fails

docs/source/notebooks/QORC-timeseries.ipynb
LF-Vigneux: Yes exactly

merlin/algorithms/layer_utils.py
LF-Vigneux: I changed the resolve_photon_loss inputs to be NoiseGroups and changed the order of computation in setup_noise_and_detectors

merlin/algorithms/layer.py
LF-Vigneux: Should be done now.

merlin/algorithms/layer_utils.py
LF-Vigneux: In your test, you check for post measurement errors but they are already implemented...

merlin/algorithms/layer_utils.py
ben9871: good

merlin/algorithms/layer.py
ben9871: done

ben9871 (approved):
changes done as requested. ready for merge

merlin/algorithms/layer_utils.py
```diff
@@ -149,13 +149,35 @@ class NoiseAndDetectorConfig:
...
+        Source noises e.g. {"indistinguishability": 0.9, "g2": 0.05, "g2_distinguishable": True}
+    circuit : dict[str, float | bool | None]
+        Circuit noises e.g. {"phase_error": 0.05, "phase_imprecision": 0.1}
+    post_measurement : dict[str, float | bool | None]
+        Post measurement noises e.g. {"brightness": 0.3, "transmittance": 0.9}
+    """
+
+    source: dict[str, float | bool | None]
+    circuit: dict[str, float | None]
+    post_measurement: dict[str, float | None]
```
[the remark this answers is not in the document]
LF-Vigneux: Adjusted in next commit

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +959,167 @@ def normalize_output_key(
...
+    elif not noise_model.phase_error == 0.0:
+        circuit["phase_error"] = noise_model.phase_error
+    if circuit == {}:
+        circuit = None
+
+    # Post-measurement noise
+    post_measurement = {
+        "transmittance": noise_model.transmittance,
+        "brightness": noise_model.brightness,
+    }
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +959,167 @@ def normalize_output_key(
...
+    if noise_model is None and empty_detectors:
+        return
+    if (noise_model is not None) and return_object:
+        raise NotImplementedError(
+            "The noise computation with the return_object feature set at True is not yet implemented."
+        )
+    if output == "amplitudes":
+        raise ValueError(
+            "When doing a noisy simulation, the probabilities measurement strategy must be used."
+        )
```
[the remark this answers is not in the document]
LF-Vigneux: But all default values return None with this line in classify_noise_model 
`if (source is None) and (circuit is None) and (post_measurement is None):
        return None`

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +959,167 @@ def normalize_output_key(
...
+    layer_noise_model: pcvl.NoiseModel | None,
+    experiment_noise_model: pcvl.NoiseModel | None,
+) -> pcvl.NoiseModel | None:
+    """Normalize the NoiseModels from two different inputs in the QuantumLayer.
+
+    Parameters
+    ----------
+    layer_noise_model: pcvl.NoiseModel | None
+        The noise model declared in the noise_model argument of the QuantumLayer.
+    layer_noise_model: pcvl.NoiseModel | None
```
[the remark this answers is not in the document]
LF-Vigneux: Done in next commit

merlin/algorithms/layer_utils.py
```diff
@@ -584,32 +612,29 @@ def resolve_circuit(
...
-        noise_model = None
-        has_custom_noise = False
     elif circuit_source.source_type == "circuit":
         if circuit_source.circuit is None:
             raise RuntimeError("Circuit must be provided for circuit source type.")
         circuit = circuit_source.circuit
         experiment = pcvl_module.Experiment(circuit)
-        noise_model = None
-        has_custom_noise = False
     elif circuit_source.source_type == "experiment":
```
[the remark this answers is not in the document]
LF-Vigneux: Fixed in next commit

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +959,167 @@ def normalize_output_key(
...
+
+    Returns
+    -------
+    pcvl.NoiseModel | None
+        The declared noise model or None if the is none.
+    """
+    if layer_noise_model is None:
+        return experiment_noise_model
+    else:
+        if experiment_noise_model is None:
```
CassNot, line 1132: I think we miss the test `normalize_noise_model(None, None)`
I understand the output is `None` and then the `QuantumLayer` treats it as `None` but it should be tested

merlin/algorithms/layer_utils.py
```diff
@@ -885,3 +959,167 @@ def normalize_output_key(
     if isinstance(key, torch.Tensor):
         return tuple(int(v) for v in key.tolist())
     return tuple(int(v) for v in key)
+
+
+def classify_noise_model(noise_model: pcvl.NoiseModel | None) -> NoiseGroups | None:
```
CassNot, line 971: The "default" value are a bit implicit and the method is not direct to understand

Do you think you could start by defining the default:

```
NOISE_GROUPS = {
    "source": {
        "indistinguishability": 1.0,
        "g2": 0.0,
        "g2_distinguishable": True,
    },
    "circuit": {
        "phase_imprecision": 0.0,
        "phase_error": 0.0,
    },
    "post_measurement": {
        "brightness": 1.0,
        "transmittance": 1.0,
    },
}
```

Then you can classify by iterating over that mapping and you can keep values that are not None and not equal to their neutral value. 

That could make the behavior visible, extensible, and easier to test.

merlin/measurement/photon_loss.py
```diff
@@ -346,6 +349,63 @@ def to(self, *args, **kwargs):
 
 
 def resolve_photon_loss(
+    noise_groups: NoiseGroups | None, n_modes: int
```
CassNot, line 352: Same problem here, the behaviour is a bit tricky to understand

And
- the docstring still mentions the photon loss is resolved from the experiment's noise model when it is now from the noise group
- `empty_noise_model` is a bit counterintuitive because it is True if there is circuit noise but no brightness/transmittance. Maybe you could call it `empty_post_measurement` or 'empty_photon_loss' ? maybe you could also add a test for this semantic issue if there is circuit noise

CassNot, line 352: Lack of tests for this method: some examples
- `noise_groups=None` returns `([1.0] * n_modes, True)`
- `NoiseGroups(..., post_measurement=None)` returns `([1.0] * n_modes, True)`
- `brightness` only returns `([brightness] * n_modes, False)`
- `transmittance` only returns `([transmittance] * n_modes, False)`
- `brightness` and `transmittance` returns `([brightness * transmittance] * n_modes, False)`

tests/algorithms/test_noise_contract.py
CassNot, line 1: I would add another test to test brightness + transmittance survival. You can verify that layer._photon_survival_probs makes sense on an empty circuit or HOM circuit for example

CassNot (changes requested):
The documentation builds correctly !
The edits for the kernels are here to keep the backend separate for now and will be edited later
Some tricks, imprecision and missing tests but otherwise, good job !

merlin/algorithms/layer_utils.py
LF-Vigneux: Added tests/algorithms/noise_contract.py::test_no_noise in next commit

merlin/algorithms/layer_utils.py
LF-Vigneux: Added comments in next commit

merlin/measurement/photon_loss.py
LF-Vigneux: Added tests/measurement/test_photon_loss.py::test_resolve_photon_loss_with_noise_groups_empty_post_measurement and modified the parameter name and docstring

LF-Vigneux: Added tests/measurement/test_photon_loss.py::test_resolve_photon_loss_with_noise_groups_empty_post_measurement

CassNot (approved):

CassNot (approved):
Strong plumbing that should encourage clean implementation of noise models

PR #209, opened 2026-05-08
PML-280: introducing the ReservoirClassifier

CassNot:
## Summary

I am introducing the merlin.models for easier QORC experiments: the ReservoirClassifier.

## Related Issue
   
PML-280

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

The main API is as follow

```
# defining the classifier
reservoir = ReservoirClassifier(in_features, 
                    out_features, 
                    n_photons, 
                    reduction=None, 
                    concatenate=True, 
                    cache=True, 
                    seed=None, 
                    device=None, 
                    dtype=None)

#fits optional sklearn decomposition, input scaling, reservoir features, feature normalization stats, and initializes the readout
reservoir.fit_reservoir(X_train)

# converts raw input into readout-ready TensorDataset
train_dataset = reservoir.make_dataset(X_train, y_train)

# applies only the trainable readout to already-computed readout features
reservoir(x)

# a training epoch can therefore look like
logits = reservoir(features)
loss = criterion(logits, targets)
loss.backward()
optimizer.step()

# runs raw inputs through preprocessing, frozen reservoir, and readout, returning logits
reservoir.predict(X_train)
```

Additionally, `save(path)` / `load(path, map_location=None, device=None)` serialize and restore model configuration, frozen reservoir state, preprocessing stats, cache, and readout weights

Layer configuration:

- `reservoir.layer.n_modes = ... `rebuilds the reservoir with a new mode count.
- `reservoir.layer.measurement_strategy = ...` rebuilds with a custom measurement strategy.
- `reservoir.layer.noise_model = ...` rebuilds with a Perceval noise model.
- `reservoir.layer.output_size`  exposes the current quantum feature dimension.
- `reservoir.processor` sets the processor (can be a MerlinProcessor for cloud processing) experiments

## How to test / How to run

1. Command lines

```
pytest tests/models/
```

I am also adding a demonstration notebook (ReservoirClassifier.ipynb)

## Documentation
- [x] Docstrings updated
- [x] Updated the API

## Checklist
- [x] Lint passes (ruff)
- [x] With this command:  `SPHINXOPTS="-W --keep-going -n" make -C docs clean html`, the docs are built without any warning or errors.


merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,988 @@
...
+        self.dtype = resolved_dtype
+        # The reduction needs to be validated and stored before inferring
+        # the logical encoded feature width.
+        self._reduction_template = self._validate_reduction(reduction)
+        self.reduction = (
+            clone(self._reduction_template)
+            if self._reduction_template is not None
+            else None
+        )
+        self.encoded_input_features = self._infer_encoded_input_features()
```
ben9871, line 196: The ticket requires the circuit width to be pca_components + 1 when a reduction is used, and in_features + 1 otherwise. 
This implementation sets quantum_input_features equal to encoded_input_features, so PCA(n_components=11) builds an 11-mode circuit instead of the required 12-mode reservoir. 
This also makes valid configurations fail, for example n_photons == n_components + 1.

merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,988 @@
...
+        reduced_normalized = self._normalize_min_max(
+            reduced,
+            self._input_min,
+            self._input_max,
+        )
+        # 4. Encode the normalized features into quantum states using the reservoir's quantum layer.
+        quantum = self._encode_quantum(reduced_normalized, processor=processor)
+        self._quantum_mean = quantum.mean(dim=0).detach().cpu()
+        self._quantum_std = quantum.std(dim=0, unbiased=False).detach().cpu()
+        # 5. Cache the quantum features of the training data if caching is enabled, and store a fingerprint of the input data for cache validation during prediction.
```
ben9871: The ticket says cache=False should make the reservoir fit step fit the PCA/scaler and skip the quantum pass, while make_dataset() re-encodes on demand.
 The current fit_reservoir() path calls _encode_quantum() unconditionally and then discards the result when self.cache is false. That still pays the full simulator or QPU cost during sampling, and processor.forward() is called even in the memory-saving path. 

This would be ok if fit_reservoir was usable on its own but what is being returned is just a model instance.
If I want to reuse the quantum embeddings for another algorithm that is not classification then I must use private methods such as `_encode_quantum` or `_fit_quantum_cache`(which is only available when caching is turned on). Private methods should also not be documented or user-facing options otherwise. For me, it should be easy for the user to do fit_reservoir and then go off and do another algorithm if needed, using the embeddings. 

So I think to justify this recomputation, you should make fit_reservoir have more value in and of itself.

merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,988 @@
...
+        ValueError
+            If ``in_features``, ``out_features``, or ``n_photons`` is not a
+            strictly positive integer, or if ``reduction.n_components`` is not
+            a strictly positive integer.
+        TypeError
+            If ``reduction`` is not ``None`` and does not behave like a
+            scikit-learn decomposition estimator.
+        """
+        super().__init__()
+
```
ben9871, line 171: This block only checks positivity. It then converts values to int later if necessary. However, a value like '4.5' photons or '3.2' features should simply not succeed. We should have value errors here for positivity and being an integer without silently accepting these values.

ben9871 (changes requested):
1. N modes = N features instead of N+1 as in the ticket
2. non int values are silently accepted by the constructor and converted to int later
3. non cache route recomputes quantum encodings and the user only has private methods(i.e undocumented) to inspect the quantum embeddings if they want to reuse them for another algorithm. I recommend making some methods public API for retrieving the embeddings.

merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,988 @@
...
+            clone(self._reduction_template)
+            if self._reduction_template is not None
+            else None
+        )
+        # 3. Normalize the reduced features to a [0, 1] range based on the min and max of the reduced data.
+        reduced = (
+            np.asarray(self.reduction.fit_transform(X_np), dtype=np.float32)
+            if self.reduction is not None
+            else X_np
+        )
```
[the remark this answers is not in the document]
CassNot: Agreed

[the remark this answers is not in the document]
CassNot: `fit_reservoir() `now calls `reduction.fit(X_np)` followed by `reduction.transform(X_np)`

And added tests with a fake estimator that has `fit` and `transform` but no `fit_transform`

merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,988 @@
...
+
+    @staticmethod
+    def _fingerprint(X: np.ndarray) -> str:
+        # Build a cheap cache key for X without storing the full raw dataset:
+        # we hash shape/dtype plus a small strided subsample, which is enough
+        # to detect the common "same training set reused" case.
+        stride = max(1, len(X) // 1000)
+        sample = np.ascontiguousarray(X[::stride][:1000])
+        payload = f"{X.shape}:{X.dtype}".encode() + sample.tobytes()
+        return hashlib.md5(payload, usedforsecurity=False).hexdigest()
```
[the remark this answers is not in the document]
CassNot: This is true. Since MD5 (and SHA1) are cryptographically broken, Bandit warns on any use of them, regardless of context. It has no way to know whether the hash is securing a password or just keying a dict.
I am therefore using `blake2b`

docs/source/notebooks/token_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+Recommended .env format:
+    CLOUD_TOKEN=your_token_here
+"""
+
+import os
+from pathlib import Path
+
+from dotenv import load_dotenv
+
```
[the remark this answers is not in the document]
CassNot: I do not want to overflow the required env with convenient files, so I just added a comment `# You can install it with: pip install python-dotenv`

merlin/models/reservoir_classifier.py
CassNot: I agree. However, there are a few things to take into account:

- `fit_reservoir()` currently needs the quantum pass to compute `_quantum_mean` /` _quantum_std`, 
- it is because later, `make_dataset()` standardizes quantum features using train-set reservoir statistics. 
- So if `cache=False` skips the quantum pass, we need to move that train-stat computation somewhere else, not just delete the call

Therefore, the fix should include the public transform method (otherwise users are still pushed toward private internals)

merlin/models/reservoir_classifier.py
CassNot: Right ! I tightened the constructor validation

merlin/models/reservoir_classifier.py
```diff
@@ -0,0 +1,1063 @@
...
+    def _rebuild_quantum_layer(self) -> None:
+        processor = getattr(self._quantum_layer, "processor", None)
+        self._quantum_layer = self._build_layer(
+            unitary_matrix=self._unitary_matrix,
+            encoded_input_size=self.encoded_input_features,
+            n_modes=self.quantum_input_features,
+            measurement_strategy=self._measurement_strategy,
+            noise_model=self._noise_model,
+        )
+        self._quantum_layer.processor = processor
```
CassNot, line 452: Breaks contract

LF-Vigneux:
Claude spotted these problems in my release highlights PR

2. Duplicate merlin.models entry in pyproject.toml

packages = [
    ...
    "merlin.models",        # already present above
    "merlin.measurement",
    "merlin.models",        # <-- duplicate added in this PR
    ...
]
The new line is a duplicate of the existing "merlin.models" entry. Remove the duplicate. [Fix this →](https://claude.ai/code?q=In%20pyproject.toml%20on%20branch%20PML-362-release-highlights-notebook-0.4%2C%20remove%20the%20duplicate%20%22merlin.models%22%20entry%20added%20to%20the%20packages%20list.&repo=merlinquantum/merlin)

3. python-dotenv is not a declared docs dependency
docs/source/notebooks/token_utils.py does from dotenv import load_dotenv, and it's imported by ReservoirClassifier.ipynb and QORC-demo-MNIST.ipynb. dotenv isn't listed in requirements-docs.txt or docs/requirements.txt. If these notebooks are executed during the Sphinx build (which runs with -W --keep-going), the missing dependency will break the build. Please add python-dotenv to the docs requirements (or guard the import).

🟡 Worth a look
4. Unintended version changes in .pre-commit-config.yaml
ruff is downgraded v0.15.12 → v0.15.11 while mypy is bumped v1.16.0 → v1.20.2. The ruff downgrade looks accidental — please confirm it's intentional.

5. Exqalibur docs link downgraded in aliases.rst
The StateVector external link was changed v1.2 → v1.1:

https://perceval.quandela.net/docs/v1.1/reference/exqalibur/index.html
Since this release line moved to perceval>=1.2, this should likely stay v1.2.

6. token_utils.py minor concerns

Hardcoded path Path("Quantum_Machine_Learning-MerLin/.env") in _ENV_SEARCH_PATHS looks like a leftover from a local dev layout — unlikely to be meaningful inside this repo.
Uses print(...) for diagnostics instead of warnings/logging; fine for a demo helper but noisy in a rendered notebook.

LF-Vigneux (approved):
Let's merge!

CassNot:
@claude review

ben9871 (approved):
Good to go

CassNot:
@ben9871 if my last commit passes test on GPU, it is ready for merging

PR #212, opened 2026-05-12
PML-319: bugfix for the loss of the Builder GPU test

CassNot:
## Summary

There was a bugged GPU test for the builder. The loss of `test_entangling_layer_models_on_gpu` was logits.sum(), which is always 1 and therefore the gradients are almost null and the test was failing.
The tests started failing after tag 2.3.
The commit 3e9b3df changed default FOCK to UNBUNCHED. As UNBUNCHED is normalized, the loss is a constant and the gradient of the constant is null. When it was FOCK, because FOCK is derived from the AMPLITUDES and not normalized, even if the loss is close to 1, it was not exactly 1 (so the gradient was not completely null)

## Related Issue
   
PML-319 

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

The loss is now the difference between the input and the logits (it does not make physical sense but the gradients are non null)

## How to test / How to run

1. Command lines

```
Block of code
```

## Documentation
- [ ] Docstrings updated
- [ ] Updated the API

## Checklist
- [x] Lint passes (ruff)
- [x] With this command: 


tests/builder/test_circuit_builder.py
LF-Vigneux, line 1: Remove the copilot comment

LF-Vigneux (changes requested):
Just remove the LLm comment and everything is good!

tests/builder/test_circuit_builder.py
CassNot: I updated the comment to say that we want the loss to vary (non-trivially)

LF-Vigneux (approved):
The bug is indeed fixed

PR #215, opened 2026-05-15
PML-309: FidelityBackend to use the QuantumLayer as a shared backend

CassNot:
## Summary

This PR refactors `FidelityKernel` to delegate kernel computation to an internal `_CCInvQuantumLayer` backend built on top of `QuantumLayer`. The goal is for the Kernel and the QuantumLayer to use the same backend (and not different ones as it is the case today) so any change in the backend of the QuantumLayer should be directly applied to the Kernel. This would be useful for noise management for instance.

`FeatureMap` is now used as a descriptor for circuit, experiment, input, dtype, and device metadata. The legacy `FeatureMap.compute_unitary` path is kept for compatibility, but is marked deprecated and no longer used by `FidelityKernel`. I do not think these methods were directly used by users but I prefer to have them with deprecations warnings first, just in case.

All existing tests path and new tests were added to validate the implementation.

## Related Issue
   
PML-309

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- Add `_CCInvQuantumLayer` as the internal backend for `FidelityKernel`.
- Move encoding, unitary construction, SLOS probability computation, photon loss, and detector transforms into the backend layer.
- Register trainable kernel parameters through the backend module.
- Keep `FeatureMap.compute_unitary` as a deprecated legacy compatibility path.
- Update deprecation registry for `FeatureMap.compute_unitary`.
- Update kernel tests for backend behavior, gradients, trainable parameters, scalar/matrix kernels, deprecations, and Perceval SLOS comparisons.
- Update measurement tests to access detector transforms through `kernel._quantum_layer`.
- Update kernel API docs, expert docs, notebook example, gallery metadata, and Iris kernel image.

## How to test / How to run

1. Command lines

```
tests/algorithms/test_kernels.py tests/algorithms/test_kernels_deprecation.py
tests/measurement/test_detectors.py::TestDetectorsWithKernels
tests/measurement/test_photon_loss.py::TestPhotonLossWithFidelityKernel
```

No benchmark was run but the training Gram matrix path keeps the upper-triangle optimization to avoid duplicate pair simulations.

## Documentation
- [x] Docstrings updated
- [x] Updated the API

## Checklist
- [x] Lint passes (ruff)
- [x] With this command:  `SPHINXOPTS="-W --keep-going -n" make -C docs clean html` the docs are built without any warning or errors.


CassNot:
@codex review focus parity and maintainability between FidelityKernel and QuantumLayer

merlin/algorithms/kernels.py
```diff
@@ -800,106 +1331,37 @@ def __init__(
...
-            ) from exc
-
-        self._detector_transform = DetectorTransform(
-            self._photon_loss_keys,
-            self._detectors,
+        self._quantum_layer = _CCInvQuantumLayer(
+            experiment=self.experiment,
+            input_state=input_state,
+            input_size=self.input_size,
+            input_parameters=[self.feature_map.input_parameters],
```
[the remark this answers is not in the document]
ben9871: yes I will second this. The angle encoding scale is ignored. This construction drops the builder angle-encoding metadata before the CCInv backend can use it. FeatureMap.simple(..., angle_encoding_scale=2.0) records the scale in feature_map._angle_encoding_specs, but this call only passes experiment=, input_parameters, and trainable_parameters. Since the backend is then initialized as an experiment source, QuantumLayer resolves angle_encoding_specs={} so scale 1.0 and 2.0 produce identical kernel values with fixed trainables.

merlin/algorithms/kernels.py
```diff
@@ -661,10 +849,314 @@ def build_fidelity_kernel(
...
+        x = x.to(dtype=self.dtype, device=self.device).reshape(-1)
+        prefix = self.input_parameters[0] if self.input_parameters else None
+
+        # Angle encoding specs (from CircuitBuilder) take priority.
+        if prefix and self.angle_encoding_specs:
+            spec = self.angle_encoding_specs.get(prefix)
+            if spec:
+                return self._prepare_input_encoding(x, prefix)
+
+        # No CircuitBuilder metadata available: the circuit was constructed
```
ben9871, line 1122: This checks for angle_encoding_specs and if not, assumes a pcvl.Circuit encoding. I think this makes errors in angle_encoding_specs within simple() less observable. The simple has this api and always has angle_encoding_specs thus an absence of angle_encoding_specs/scaling should not be silently accepted but raise an error( as it is never directly constructed like a perceval circuit ).

```python
 def simple(
        cls,
        input_size: int,
        *,
        dtype: str | torch.dtype = torch.float32,
        device: torch.device | None = None,
        angle_encoding_scale: float = 1.0,
        n_modes: int | None = None,
    ) -> "FeatureMap":
```

ben9871 (changes requested):
Mostly good, just a slight issue with angle_encoding_specs not being passed through the pipeline that codex also indicated. And a change to fallbacks to make sure outright errors don't silently pass for simple()

merlin/algorithms/kernels.py
```diff
@@ -942,44 +1413,26 @@ def forward(
...
-                    .conj()
-                    .to(x1.device)
-                    for x in x2
-                ])
-            else:
-                raise (TypeError("x2 is not None nor torch.Tensor"))
-
-            # Calculate circuit unitary for every pair of datapoints
+            # All (len_x1 × len_x2) pair unitaries
             all_circuits = U_forward.unsqueeze(1) @ U_adjoint.unsqueeze(0)
```
LF-Vigneux: Why does the forward of _CCInvQuantumLayer not handle this composition. I think that the FidelityKernel should only call layer(x_1,x_2) to create its kernel  mt=atrix and it is the _CCInvQuantumLayer's responsability to create the unitary and compute the probabilities _compute_transition_probs should compute the unitaries.

merlin/algorithms/kernels.py
```diff
@@ -1077,35 +1506,14 @@ def _return_kernel_scalar(
...
-        if self._input_detection_index is not None:
-            value = detection_probs[0, self._input_detection_index]
-        else:
-            weights = self._input_detection_weights.to(
-                dtype=detection_probs.dtype, device=detection_probs.device
-            )
-            value = (detection_probs @ weights)[0]
-
-        return value.item()
+        return transition_probs[0].item()
```
LF-Vigneux: Again the _compute_transition_probs should compute the kernel unitary

LF-Vigneux (changes requested):
Different role attribution. Also, for this point in the ticket:

- computation_process.converter is restored to the original after each forward() call.

I do not see where it is done.

merlin/algorithms/kernels.py
```diff
@@ -661,10 +703,314 @@ def build_fidelity_kernel(
...
+        input_parameters: list[str],
+        trainable_parameters: list[str],
+        computation_space: ComputationSpace,
+        dtype: torch.dtype,
+        device: torch.device,
+    ) -> None:
+        super().__init__(
+            experiment=experiment,
+            input_state=input_state,
+            input_size=input_size,
```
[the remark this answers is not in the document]
CassNot: This failure case will be implemented in the QuantumLayer as well. There should not be silent ignore of inconsistent `input_size` / `input_parameters`. So this is intentional for the new `FidelityKernel` backend: encoder is legacy-only and is only used by deprecated `FeatureMap.compute_unitary()`. Kernel computation now requires exact input-size compatibility or builder-provided angle-encoding metadata. 

However, I am making the failure explicit and updating docs/tests in this sense !

merlin/algorithms/kernels.py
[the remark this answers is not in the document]
CassNot: the code wires the `angle_encoding_specs=self.feature_map._angle_encoding_specs` to `FidelityKernel and therefore _CCInvQuantumLayer`
But I added regression tests so this cannot silently regress

-     test_kernel_backend_preserves_builder_angle_encoding_scale
-     test_simple_kernel_backend_preserves_angle_encoding_scale
-     test_kernel_backend_preserves_builder_subset_scale

merlin/algorithms/kernels.py
CassNot: Agreed, I implemented a fix:

- Builder/simple FeatureMap instances now record that angle-encoding metadata is required (and FidelityKernel validates required builder metadata at construction)
- _CCInvQuantumLayer._encode_single() now raises if required builder metadata is missing instead of falling through to direct pcvl.Circuit compatibility behavior
- Added validation for missing combinations and missing scales
- Direct pcvl.Circuit / pcvl.Experiment feature maps still keep the compatibility fallback

merlin/algorithms/kernels.py
CassNot: I am not sure. `_compute_transition_probs` takes already-composed circuit unitaries and runs SLOS + photon loss + detector transforms. 
I think that separation is useful for tests and debugging -> the Perceval reference tests call _compute_kernel_unitary() and then _compute_transition_probs() specifically to isolate “given this unitary, do we compute the right transition probability?”
What is your opinion on that @LF-Vigneux ?

merlin/algorithms/kernels.py
CassNot: I agree and implemented the changes
`_CCInvQuantumLayer.forward()` now does:

- batch unitary construction
- pairwise composition U(x1) @ U†(x2)
- transition probability computation
- kernel matrix construction
- PSD projection (if requested)

`FidelityKernel.forward()` only normalizes public inputs, handles scalar/datapoint return shape, and delegates to self._quantum_layer(...)

`_compute_transition_probs()` are kept as the low-level SLOS/detector transform for already-composed unitaries (because existing backend tests use it directly for Perceval comparisons).

I also updated the docs to describe _CCInvQuantumLayer as the kernel-matrix backend

And I added a regression test that `_quantum_layer.forward` matches `FidelityKernel.forward` (and it passes)

CassNot:
@LF-Vigneux , my "computation_process.converter is restored to the original after each forward() call" was a bit wrong because no part handles a restore because `forward` does not replace the converter (but line 1228, `_compute_unitary` directly calls `self.computation_process.converter.to_tensor(*self.thetas, x_enc)`, then line 1315, batched unitary and matrices are built). So the converter is preserved across forward() calls because _CCInvQuantumLayer.forward does not mutate nor swap it

I implemented the changes, updated the changes and also added a mypy fix on the signature of forward : `_CCInvQuantumLayer.forward()` should match `QuantumLayer.forward()`’s variadic signature, and store `force_psd` on the backend instead of passing it as an extra keyword argument. Now `_CCInvQuantumLayer.forward` matches `QuantumLayer.forward`’s override shape (inputs + shots, sampling_method, and simultaneous_processes) and:

- `force_psd` is stored on the backend layer state
- `FidelityKernel.forward` now calls `layer(x1)` or `layer(x1, x2)` directly

The branch is up to date with release/0.4

@codex review my latest commits and ensure ease of future maintainability of deprecated methods

merlin/algorithms/kernels.py
LF-Vigneux: After the code refactoring, I agree

LF-Vigneux (approved):
My requested changes are there

ben9871 (approved):
Review comments fixed

PR #219, opened 2026-05-21
PML-327: Backprop guard implemented for memristor

LF-Vigneux:
## Summary
Detaching history memristor states after num_backprop_steps forward passes to limit the memristive history depth accessed by the backwards.

## Related Issue
PML-327

## Type of change
- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes
Adding a ``num_backprop_steps``parameter to the add_memristve_phase_shifter method. This argument detaches any tensor in the memristive history that is associated with a forward pass at least ``num_backprop_steps+1`` iterations away.

## How to test / How to run
```
pytest -q
```

## Documentation
- [x] Docstrings updated
- [x] Updated the API


merlin/algorithms/layer.py
```diff
@@ -1099,6 +1113,21 @@ def forward(
                             Memristive phase-shifter analyzed: {self._memristive_metadata[i]}
                         """
                     )
+
+                # Detach history that is too long ago
```
ben9871: This detaches an old tensor stored in `memristive_history`, but it does not detach the recurrent state chain used by future forwards. By the time this line runs, newer `self.memristive_state[i]` tensors have already been computed from the original graph. Replacing an old list entry with `old.detach()` does not rewrite the graph inside later states. With `num_backprop_steps=0`, a loss on the fourth forward still produces gradients for the first three inputs. The fix should detach the active state that will be reused by the next timestep, not only the historical copy kept for inspection. Concretely, the recurrence needs a truncated-BPTT(Backprop through time) boundary such that `self.memristive_state[i]` is detached when it falls outside the requested sliding window.

tests/algorithms/test_layer.py
```diff
@@ -2342,3 +2349,241 @@ def counted_update(state: torch.Tensor, output: torch.Tensor) -> torch.Tensor:
...
+    # and that the memristive state can participate in gradients
+    assert loss.grad_fn is not None
+
+    # Verify that only the last num_backprop_steps states in the history are attached
+    # to the computation graph (i.e., have requires_grad=True), while older states are
+    # detached and do not participate in backpropagation
+    num_backprop_steps_a = metadata_a["num_backprop_steps"]
+    num_backprop_steps_b = metadata_b["num_backprop_steps"]
+
+    # Check memPS_a history: only the last (num_backprop_steps + 1) should be attached
```
ben9871: This checks `requires_grad` on tensors stored in the history list, but that is not a sufficient test of the sliding-window contract. A detached old history entry can coexist with newer state tensors that still retain graph links to older forwards. Add a behavioural test that backpropagates a loss from the last timestep and verifies which previous timestep inputs/outputs receive gradients.

merlin/algorithms/layer.py
```diff
@@ -1099,6 +1113,21 @@ def forward(
...
+                position_to_detach = (
+                    len(self.memristive_history[i])
+                    - self._memristive_metadata[i]["num_backprop_steps"]
+                    - 1
+                )
+                if position_to_detach >= 0:
+                    self.memristive_history[i][position_to_detach] = (
+                        self.memristive_history[i][position_to_detach].detach()
+                    )
+                    print(self.memristive_history[i])
```
ben9871: : ) This debug `print()` runs inside `QuantumLayer.forward()` for memristive layers. It will pollute training logs and notebooks once a sequence has multiple timesteps.

merlin/builder/circuit_builder.py
```diff
@@ -476,6 +476,7 @@ def add_memristive_ps(
         update_rule: Callable,
         initial_state: float,
         name: str | None = None,
+        num_backprop_steps: int = 0,
```
ben9871: `num_backprop_steps` is a window length, so it should be validated as a non-negative integer at the builder boundary. Right now `num_backprop_steps=-1` is accepted and stored in metadata, then the forward-time indexing logic can compute an invalid position. Invalid API input should fail early with a clear `ValueError`.

merlin/algorithms/layer.py
```diff
@@ -1767,4 +1796,6 @@ def reset(self, batch_size: int = 1) -> None:
                 dtype=self.dtype,
             )
             self.memristive_history[i] = [self.memristive_state[i]]
+            if self._memristive_metadata[i]["num_backprop_steps"] == 0:
```
ben9871: This `.detach()` call has no effect as written because `Tensor.detach()` is not in-place. It returns a new detached tensor, and that return value is immediately discarded here. The reset tensor is newly created with `torch.full(...)` and normally has no graph anyway, so this branch can probably be removed. If the intent is to detach explicitly, assign the return value back: `self.memristive_history[i][0] = self.memristive_history[i][0].detach()`.

ben9871 (changes requested):
A stray debug print remains, some issues with detaching tensors and observability, can enhance tests too

ben9871 (approved):
files changed and ambiguity solved by adding a utility method for TBPTT

PR #220, opened 2026-05-21
PML-282 Photonic Generator

ben9871:
# PML-282 Add QuantumLayer-backed PhotonicGenerator

## Summary

This PR adds a `PhotonicGenerator` model that composes one or more existing `QuantumLayer` objects into a latent-to-sample generator. It keeps quantum execution inside `QuantumLayer`, exposes raw per-layer measurements for inspection, and uses output adapters to map measurements to vectors or images.

## Related Issue

Related Jira: PML-282

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- Add `merlin.models.PhotonicGenerator`.
- Add `GeneratorMeasurements` to carry raw per-layer outputs and output-key metadata.
- Add latent distribution helpers:
  - `LatentDistribution`
  - `NormalLatent`
- Add output adapter API:
  - `OutputAdapter`
  - `VectorAdapter`
  - `ImageAdapter`
- Export the new public objects from `merlin.models` and top-level `merlin`.
- Add tests covering construction, validation, latent sampling, generation, adapter behaviour, parameter registration, and gradient flow.
- Add the new API page under `docs/source/api_reference/api/merlin.models.photonic_generator.rst`.

API shape:

```python
generator = ml.PhotonicGenerator(
    layers=[layer_0, layer_1],
    output_adapter=ml.ImageAdapter(shape=(1, 8, 8)),
)

z = generator.sample_latent(batch_size=16)
samples = generator(z)
measurements = generator.measure(z)
first_layer = generator[0]
```

## How to test / How to run

1. Run pre-commit on the touched Python files:

```bash
python -m pre_commit run --files merlin/models/photonic_generator.py merlin/models/__init__.py merlin/__init__.py tests/models/test_photonic_generator.py
```

2. Run the focused test file:

```bash
python -m pytest tests/models/test_photonic_generator.py -q
```

3. Run the strict docs build:

```bash
python -m sphinx -b html -W --keep-going -n docs/source docs/build/html
```

Observed locally:

```text
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
mypy.....................................................................Passed

16 passed

build succeeded.
```

N/A.

The generator evaluates its underlying `QuantumLayer` heads sequentially and then adapts the collected outputs. Runtime is therefore approximately the sum of the runtimes of the configured quantum layers. No cross-layer batching or caching is introduced in this PR.

This is intentional for the first version because different generator heads may use different circuits, input states, output spaces, measurement strategies, and noise models.

## Documentation

- [x] Docstrings updated
- [x] Updated the API

## Checklist

- [x] Lint passes (ruff)
- [x] With this command:


merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+
+        Parameters
+        ----------
+        batch_size : int
+            Number of latent vectors to sample.
+        device : torch.device | None
+            Device on which the tensor is created. If omitted, CPU is used.
+        dtype : torch.dtype | None
+            Floating dtype of the returned tensor. If omitted, the current
+            PyTorch default dtype is used.
```
LF-Vigneux, line 160: Shouldn't the same dtype and device docstring be used for normal latent and latent? Just from a uniformity point of view

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+            _resolve_measurement_kind(layer.measurement_strategy)
+            is MeasurementKind.AMPLITUDES
+        ):
+            raise ValueError(
+                "PhotonicGenerator does not support amplitude-output layers; "
+                f"layers[{index}] uses MeasurementStrategy.amplitudes()."
+            )
+
+    latent_dim = validated_layers[0].input_size
+    for index, layer in enumerate(validated_layers[1:], start=1):
```
LF-Vigneux, line 791: Wouldn't this fail if there is only one layer

LF-Vigneux (changes requested):
The form and structure of the code seems fine to me. This is not a complete review pending the remaining functions to implement.

merlin/models/photonic_generator.py
CassNot, line 1: Maybe you could add the reference to the orignal work: https://opg.optica.org/opticaq/fulltext.cfm?uri=opticaq-2-6-458

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+
+        Raises
+        ------
+        NotImplementedError
+            If the subclass does not implement adaptation.
+        """
+        raise NotImplementedError
+
+
+class VectorAdapter(OutputAdapter):
```
CassNot, line 224: Both classes `VectorAdapter` and `ImageAdapter` do not preserve the original QGAN mapping
Here: they flatten the outputs and crop/pad them directly, ignoring the output_keys
In the original work: the orignal `PatchGenerator` regroups Fock states through output_map and sums collision with index_add_ (filters lossy/non-PNR states + renormalizes) then maps to pixels -> see method `dist_to_image` in https://github.com/merlinquantum/reproduced_papers/blob/main/papers/photonic_QGAN/lib/generators.py

Therefore, in this implementation, the mapping depends on the MerLin's combinadics and not the paper mapping (for which the code is the same in the reproduced_papers repo as in the original work, in utils/mappings.py)

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+            raise ValueError(
+                "All measurement tensor outputs must share the same batch size; "
+                f"expected {batch_size}, got {output.shape[0]} at outputs[{index}]."
+            )
+        flattened.append(output.reshape(output.shape[0], -1))
+
+    return flattened
+
+
+def _center_crop_or_pad(x: torch.Tensor, size: int) -> torch.Tensor:
```
CassNot, line 830: Here, you apply cropping and padding after concatenating all heads
But the orignal work applied it per batch before concatenation (line 173 in https://github.com/merlinquantum/reproduced_papers/blob/main/papers/photonic_QGAN/lib/generators.py)

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+    std : float
+        Standard deviation of the normal distribution. Default is ``1.0``.
+
+    Raises
+    ------
+    ValueError
+        If ``std`` is not positive.
+    """
+
+    def __init__(self, dim: int, mean: float = 0.0, std: float = 1.0) -> None:
```
CassNot, line 136: Default latent sampling does not match the reproduction training loop (line 142 https://github.com/merlinquantum/reproduced_papers/blob/main/papers/photonic_QGAN/lib/qgan.py)
PhotonicGenerator.generate() defaults to `NormalLatent(..., std=1.0)`, while the original QGAN samples latent inputs with std `2 * pi`

tests/models/test_photonic_generator.py
CassNot, line 1: A regression test should also compare a fixed set of output_keys and synthetic probability tensors against the original regroup/crop/normalize behavior

CassNot (changes requested):
There are some discrepancy with the orignal code 

- (orignal perceval) https://github.com/Quandela/photonic-qgan
- (new MerLin) https://github.com/merlinquantum/reproduced_papers/tree/main/papers/photonic_QGAN
I am describing these discrepancies in review comments

**More of a product question:** What happens if the subgenerators are all the same ? Do we need to create multiple times the same one or should there be a shortcut to have to only define it once, then through the API, we precise the circuit/layer structure and N ones would be initiated ? (as in the original one)

**Lack of validation**: to validate this implementation, I would test it in the same settings as the original work (implementation here https://github.com/merlinquantum/reproduced_papers/tree/main/papers/photonic_QGAN) in a dedicated notebook

CassNot:
@codex review with respect to original implementation https://github.com/merlinquantum/reproduced_papers/tree/main/papers/photonic_QGAN

merlin/models/photonic_generator.py
ben9871: will do in next commit

merlin/models/photonic_generator.py
ben9871: will be changed in next commit

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+        ------
+        TypeError
+            If any measurement output is not a tensor.
+        ValueError
+            If no outputs are provided, if tensor outputs are not batched, or if
+            batch dimensions differ.
+        """
+        flattened = _flatten_tensor_outputs(measurements.outputs)
+        combined = torch.cat(flattened, dim=1)
+        return _center_crop_or_pad(combined, self.size)
```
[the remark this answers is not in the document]
ben9871: correct

merlin/models/photonic_generator.py
```diff
@@ -0,0 +1,672 @@
...
+        super().__init__()
+        validated_layers = _validate_layers(layers)
+        self.layers = nn.ModuleList(validated_layers)
+        if not isinstance(output_adapter, nn.Module):
+            raise TypeError("output_adapter must be a torch.nn.Module.")
+        self.output_adapter = output_adapter
+
+        inferred_dim = validated_layers[0].input_size
+        if latent is None:
+            latent = NormalLatent(inferred_dim)
```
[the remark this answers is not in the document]
ben9871: correct, will be changed soon

merlin/models/photonic_generator.py
ben9871: correct

tests/models/test_photonic_generator.py
ben9871: correct, will do in next commit

merlin/utils/grouping.py
CassNot, line 1: This grouping strategy requires the keys and therefore breaks the `nn.Sequential` pipeline we can have with the other ones
Maybe you can move it to `models/photonic_generator.py` for now as we know it works for the GAN but we have not tested it on other models
This way, we can use it externally but it does not surprise users if it breaks `nn.Sequential` pipeline

CassNot (changes requested):
We still miss:
- a demo notebook illustrating that this model works as in https://github.com/merlinquantum/reproduced_papers/tree/main/papers/photonic_QGAN with Adam optimizer
- maybe you could add the original image from the paper in the documentation to illustrate how the model works (it is quite clear with it)

CassNot:
@ben9871 there are still some validations to do within the notebook (the loss function are a bit flat so weird)
However, from this notebook, I can see that the implementation is working. I will merge this ticket in a dedicated branch so @LF-Vigneux and @alexiasalavrakos can work in parallel on this branch (further testing and scalability study)
This way, we will be sure than when the photonicQGAN arrives in release/0.4 branch, it will be ready !

CassNot (approved):
Approve but will be edited in the release/0.4-photonicQGAN dedicated branch (further validation, comparison, benchmark and scalability study)

LF-Vigneux (approved):
Approved my first quick review, forgot to do it

PR #223, opened 2026-05-28
PML-320: shared simple builder for simple circuits through QuantumLayer and FeatureMap

CassNot:
## Summary

Introduces a single shared circuit-building helper `_build_simple_circuit` so that `QuantumLayer.simple` and `FeatureMap.simple` both delegate to the same method instead of duplicating it. 
Also deprecates `FidelityKernel.simple` and the `n_modes` parameter on both `FeatureMap.simple` and `FidelityKernel.simple`

A follow-up ticket will extend this deprecation to `KernelCircuitBuilder.build_fidelity_kernel` to complete the removal of all circuit-building paths in `FidelityKernel`.

EDIT: `input_state` in `FidelityKernel` can now be inferred from the number of modes in the circuit or from a `n_photons` parameter (similar to `QuantumLayer`). This way, we can have a `FidelityKernel` with a `FeatureMap.simple` that stays simple, without having to precise the number of photons (it is not the case in `QuantumLayer`)

## Related Issue
   
PML-320

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- `layer_utils.py`: 

1. added `_build_simple_circuit(input_size, n_modes, angle_encoding_scale)`: the single source of truth for the LI_simple → angle_encoding → RI_simple topology.
2. `QuantumLayer.simple` now delegates to `_build_simple_circuit`.

- `kernels.py`:

1. `FeatureMap.simple` now delegates to `_build_simple_circuit`.
3. `FeatureMap.simple`: n_modes parameter deprecated; mode count is now always `input_size + 1` + Marked for removal in 0.5.
4. `FidelityKernel.simple`: deprecated in favour of `FeatureMap.simple` + `FidelityKernel`; duplicate DeprecationWarning suppressed when forwarding to FeatureMap.simple. Marked for removal in 0.5.
5. Fixed None-safety bug in post-select checks (post_select_fn can be None); same fix applied in layer_utils.vet_experiment and FidelityKernel._validate_experiment (due to Perceval new release).
6. `input_state` is inferred from `n_photons` if given, otherwise from the nb of modes

- `deprecations.py`: extended `sanitize_parameters` to cover the new deprecated parameters.
- `test_layer_utils.py`: new test file covering `_build_simple_circuit`.
- `test_kernels.py`, `test_kernels_deprecation.py`: updated to assert new deprecation warnings and correct behaviour.
- `Kernels.ipynb`, `binary_classification.ipynb`, `kernels.rst`: updated to use FeatureMap.simple directly and not FidelityKernel.simple (as it will be deprecated)

## How to test / How to run

1. Command lines

```
pytest tests/algorithms/test_layer_utils.py tests/algorithms/test_kernels.py tests/algorithms/test_kernels_deprecation.py tests/algorithms/test_layer.py -v
```

## Documentation
- [x] Docstrings updated
- [ ] Updated the API

## Checklist
- [x] Lint passes (ruff)


merlin/algorithms/kernels.py
```diff
@@ -684,24 +699,7 @@ def simple(
...
-        builder.add_angle_encoding(
-            modes=list(range(int(input_size))),
-            name="input",
-            subset_combinations=False,
-            scale=angle_encoding_scale,
-        )
-
-        # Trainable entangling layer after encoding
-        builder.add_entangling_layer(trainable=True, name="RI_simple")
+        builder = _build_simple_circuit(input_size, n_modes, angle_encoding_scale)
```
LF-Vigneux, line 702: I know that this function unifies the simple circuit construction. However, in a noiseless case, the last entangling layer will always cancel out in the fidelity measurement. Is the goal to keep the same circuit even though it is not optimized for a fidelity kernel? I am just laying my thoughts here.

merlin/algorithms/kernels.py
```diff
@@ -684,24 +699,7 @@ def simple(
...
 
         return cls(
```
LF-Vigneux, line 704: Here why input_parameters=None and not input_parameters="input" just like in the simple circuit definition

merlin/utils/deprecations.py
```diff
@@ -183,7 +183,29 @@ def _remove_FidelityKernel_input_state(
...
         None,
     ),
-    # FidelityKernel.simple deprecations
+    # FeatureMap.simple deprecations
+    "FeatureMap.simple.n_modes": (
+        "The number of modes is fixed to 'input_size + 1'. "
+        "Use CircuitBuilder directly if you need a different mode count.",
+        False,
+        None,
+    ),
```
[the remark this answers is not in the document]
LF-Vigneux: I agree Feature map still accepts it in the declaration, so the deprecation is not handled correctly

merlin/utils/deprecations.py
```diff
@@ -183,7 +183,29 @@ def _remove_FidelityKernel_input_state(
...
+    # FidelityKernel.simple parameter-level deprecations
+    "FidelityKernel.simple.n_modes": (
+        "The number of modes is fixed to 'input_size + 1'. "
+        "Use CircuitBuilder directly if you need a different mode count.",
+        False,
+        None,
+    ),
```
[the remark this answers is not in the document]
LF-Vigneux: Same thing here

docs/source/api_reference/api/merlin.algorithms.kernels.rst
```diff
@@ -63,17 +63,19 @@ Quickstart: Fidelity kernel in a few lines
 
     import torch
     from merlin import ComputationSpace
-    from merlin.algorithms.kernels import FidelityKernel
+    from merlin.algorithms.kernels import FeatureMap, FidelityKernel
 
     # Build a kernel where inputs of size 2 are encoded in a 4-mode circuit
```
LF-Vigneux: THis is not what is done is input_size 2 for 3 modes

docs/source/user_guide/kernels.rst
```diff
@@ -52,7 +52,7 @@ Quick Start Decision Guide
 --------------------------
 
 **"I want to quickly try quantum kernels on my data"**
-    → Use ``FidelityKernel.simple()`` with default parameters
+    → Build a feature map with ``FeatureMap.simple()`` and pass it to :class:`~merlin.algorithms.kernels.FidelityKernel`
```
LF-Vigneux: You could hyperlink it like the class but with :meth: so that the users can directly click on the method

LF-Vigneux (changes requested):
Just a couple of changes that are technicalities and a thought

merlin/algorithms/kernels.py
CassNot: That is a good question and I agree on that (and I think we had discussed it at some point)
To me, the goal is to keep the same circuit to highlight the fact that we have the same one. 
If the user calls feature_map.circuit or kernel.featuremap.circuit, it should be clear what the circuit looks like even though computation can cancel the second entangling layer after.

merlin/utils/deprecations.py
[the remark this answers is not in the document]
CassNot: Both are now changed to say that the n_modes setter is still honored in 0.4 but that this will not be handled in 0.5 anymore

merlin/algorithms/kernels.py
```diff
@@ -1795,18 +1804,21 @@ def simple(
...
+        # registry entry, so a second warning would be confusing.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            feature_map = FeatureMap.simple(
+                input_size=input_size,
+                n_modes=n_modes,
+                dtype=dtype,
+                device=device,
+                angle_encoding_scale=angle_encoding_scale,
+            )
```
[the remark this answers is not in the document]
CassNot: I believe the cleanest fix wrt the other comments above is to keep runtime warning ownership in `depreactions.py` and remove any filtering in `kernels.py`. I am centralizing them in `deprecations.py`

merlin/algorithms/kernels.py
CassNot: None tells FeatureMap.__init__ to infer the input prefix from the CircuitBuilder (line 308)
I hesitated between not specifying any or highlighting that None -> circuit builder inputs
I am adding an in-line comment to clarify

docs/source/api_reference/api/merlin.algorithms.kernels.rst
CassNot: True ! I changed the comment

LF-Vigneux (approved):
Changes good

PR #227, opened 2026-06-04
PML-328 : Bugfix QLayer manage input_state and n_photons incompatibility error

leithkarrai:
## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## How to test / How to run

To test this functionality, simply run test_layer.py file. the last 5 tests are designed to test the bug fix, so if 100% of those tests pass,
the bugfix is solved.
1. Command lines

```
cd merlin
pytest tests/algorithms/test_layer.py
```

If you initialize QuantumLayer with an incompatible number of photons and input states :
QuantumLayer(
    input_size=0,
    circuit=pcvl.Circuit(3),
    input_state=pcvl.StateVector("|1,0,1>")+pcvl.StateVector("|0,1,1>"), 
    n_photons=6, # Too many photons compared to the input_state
    measurement_strategy=MeasurementStrategy.probs(),
)
It will return this error:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/mnt/c/Users/LeïthKARRAÏ/PML-328/merlin/tests/algorithms/test_layer.py", line 2939, in <module>
    QuantumLayer(
  File "/mnt/c/Users/LeïthKARRAÏ/PML-328/merlin/merlin/utils/deprecations.py", line 535, in wrapper
    return func(*f_args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/LeïthKARRAÏ/PML-328/merlin/merlin/algorithms/layer.py", line 234, in __init__
    raise ValueError("number of photons doesn't fit input state")
ValueError: number of photons doesn't fit input state

## Documentation
- [x] Docstrings updated
- [x] Updated the API

## Checklist
- [x] Lint passes (ruff)
- [x] With this command: 

In this ticket, we ensured that if n_photons and input_state do not match (hence the number of photons is not the same), __init__ returns a ValueError. We achieved this by adding a new function in layer_utils which computes the number of photons in the input_state, regardless of its type, and we compare it in the __init__ method to check if they are identical. Therefore, you can still initialize a class if one variable is None while the other is initialized.

To test the changes, we initialized a class with incompatible variables and verified that the output is always ValueError: number of photons doesn't fit input state. All tests are located in tests/algorithms/test_layer.py.


merlin/algorithms/layer_utils.py
```diff
@@ -925,3 +925,33 @@ def normalize_output_key(
...
+
+
+def sum_input_elements(input_state) -> float | int:
+    """
+    compute number of photons from input_state.
+    """
+    if input_state is None:
+        return 0.0
+
+    # 1. Structures de données classiques (Calcul direct)
```
CassNot: The comments should be in english to improve future maintainability of the code (people who read it are not necessarily francophone)

merlin/algorithms/layer_utils.py
```diff
@@ -925,3 +925,33 @@ def normalize_output_key(
     if isinstance(key, torch.Tensor):
         return tuple(int(v) for v in key.tolist())
     return tuple(int(v) for v in key)
+
+
+def sum_input_elements(input_state) -> float | int:
```
CassNot: In what cases is your sum a float ?

tests/algorithms/test_layer.py
```diff
@@ -2889,3 +2891,48 @@ def test_long_sequence_with_manual_sliding_window_detach():
 
     for index, grad_norm in enumerate(grad_norms[: n_steps - k]):
         assert grad_norm < 1e-8, f"Input {index} should not be in the TBPTT window"
+
+def test_quantum_layer_photon_count_mismatch_tensor():
```
CassNot: Your tests should also verify that there is no ValueError when it is not expected (for instance, the case above should be working)

merlin/algorithms/layer_utils.py
```diff
@@ -925,3 +925,33 @@ def normalize_output_key(
...
+        return torch.sum(input_state).item()
+
+    if isinstance(input_state, (list, tuple)):
+        return sum(input_state)
+
+    # 2. Structures quantiques Perceval (Représentations creuses)
+    if isinstance(input_state, pcvl.BasicState):
+        return input_state.n
+
+    if isinstance(input_state, pcvl.StateVector):
```
CassNot: input_state.n if input_state is a StateVector is a set and not an integer
Therefore, even if you had same number of photons in n_photons and input_state, your test would fail. For instance, the code below fails with `FAILED tests/algorithms/test_layer.py::test_quantum_layer_accepts_matching_statevector_photon_count - ValueError: number of photons doesn't fit input state` even though it should work

```
def test_quantum_layer_accepts_matching_statevector_photon_count():
    state_vector = pcvl.StateVector("|1,0,1>")

    layer = QuantumLayer(
        input_size=0,
        circuit=pcvl.Circuit(3),
        input_state=state_vector,
        n_photons=2,
        measurement_strategy=MeasurementStrategy.probs(),
    )

    assert layer.n_photons == 2
```

CassNot (changes requested):
Good first implementation !

However, there are a few things to address before we can merge:

**Main concern**: I have 14 tests failing (they should also fail in the automatic tests but the workflow exits before because of ruff formatting issue)

Image

It is because your tests do not take into account when tensors are a batch of amplitudes (for instance in `test_compute_ebs_simultaneously.py`). Therefore, the sum on the tensor is not related to n_photons. The correct validation for tensors is shape-based, using `n_photons`, `n_modes`, and `ComputationSpace`, which the existing `_validate_amplitude_input()` / `_embed_amplitude_tensor()` path already does (which brings me to my second point). Therefore, I do not think you need to handle tensors input here (`input_state` as tensors will be deprecated to favour StateVector input)

**Second issue to address**: the method `prepare_input_state()` normalizes input_state to canonical form. If you have the following code
```
circuit = pcvl.Circuit(2)
experiment = pcvl.Experiment(circuit)
experiment.with_input(pcvl.BasicState([1, 0]))  # valid 1-photon experiment input

QuantumLayer(
    input_size=0,
    experiment=experiment,
    input_state=[1, 1],  # should be ignored by experiment.input_state
    n_photons=1,
    measurement_strategy=MeasurementStrategy.probs(),
)
```
it would fail with the new ValueError. However, in this case, `input_state` is ignored  and replaced with the `experiment.input_state` (overwrites the input_state). This scenario should not fail. One solution for that is to run your new method after `prepare_input_state`

**Third issue**: related to my comment below on statevector.n

**Documentation**: Your new method is not documented (no input type annotation and no project-format NumPy docstring). You can either make it private and typed, or document it properly and add it to the API docs if intended public

merlin/algorithms/test.py
CassNot, line 1: Ce fichier de test est toujours en trop et devrait être retiré

tests/algorithms/test_layer.py
```diff
@@ -2889,3 +2890,71 @@ def test_long_sequence_with_manual_sliding_window_detach():
 
     for index, grad_norm in enumerate(grad_norms[: n_steps - k]):
         assert grad_norm < 1e-8, f"Input {index} should not be in the TBPTT window"
+
```
CassNot, line 2913: Comments should be in english

merlin/algorithms/layer_utils.py
```diff
@@ -373,8 +373,11 @@ def prepare_input_state(
         return input_state, n_photons
 
     # === Handle tuple/list (convert to BasicState) ===
-    if isinstance(input_state, tuple):
-        input_state = list(input_state)
+    if isinstance(input_state, (list, tuple)):
+        input_list = list(input_state)
```
CassNot: Line 380:  this converts every list/tuple element with int(x)
This could hide invalid fractional occupations instead of failing. Example: [1.9, 0.1, 0] becomes BasicState([1, 0, 0]) and can pass the photon-count check. Is it expected ?
In this case, maybe we should reject fractional list instead of creating a wrong number of photons

CassNot:
```
For instance
def test_prepare_input_state_rejects_fractional_sequence_occupations():
    with pytest.raises((TypeError, ValueError)):
        prepare_input_state(
            [1.1, 0.9],
            1,
            ComputationSpace.UNBUNCHED,
            None,
            torch.complex64,
        )
```
You can fail if the input_state is a list AND made of float because

- superposed inputs (with floats : [1.9, 0.1, 0]) should be passed as StateVector 
- non-superposed inputs ([1,1,0]) could work

If the input state is a list then it should be made of integers >=0

merlin/algorithms/layer_utils.py
```diff
@@ -983,3 +928,17 @@ def normalize_output_key(
     if isinstance(key, torch.Tensor):
         return tuple(int(v) for v in key.tolist())
     return tuple(int(v) for v in key)
+
+
+def extract_photon_count(input_state) -> int | None:
```
CassNot: Add documentation (cf old comments)

CassNot: -> Doc = input type annotation and project-format NumPy docstring) (cf method abvove)
You can either make it private and typed, or document it properly and add it to the API docs if intended public

CassNot (changes requested):

CassNot:
@claude review

merlin/algorithms/layer_utils.py
```diff
@@ -983,3 +927,25 @@ def normalize_output_key(
...
```
CassNot: still missing a precision on `input_state`:
`def extract_photon_count(input_state: StateVector | pcvl.StateVector | pcvl.BasicState)`
and maybe `None` as well as you have the `if` on `None` just below

CassNot (changes requested):
Just one tiny change for consistency !

CassNot:
@claude review updates

CassNot (approved):
I solved the conflicts due to the NoisySLOS merge and I approved

PR #232, opened 2026-06-04
PML-287: circuit noise - phase error & imprecision

CassNot:
## Summary

Adds circuit phase-noise support for `QuantumLayer`/SLOS simulations:

- `phase_imprecision` now applies deterministic nearest-grid phase quantization
- `phase_error` now samples stochastic phase perturbations and averages probability distributions over Monte Carlo unitary samples.

I also fixed GPU computation for NoisySLOS and added dedicated tests for GPU coverage, ensuring all noisy computations are done on the device

## Related Issue
   
PML-287

## Type of change
- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

Phase error & imprecision implementation:
- Removes the previous `NotImplementedError` path for circuit phase noise in `QuantumLayer`
- Adds `n_phase_error_samples` to control Monte Carlo averaging when active `phase_error` is present
- Routes `phase_imprecision` and `phase_error` from `NoiseModel` through `NoiseGroups`, `QuantumLayer`, `ComputationProcess`, and `CircuitConverter`
- Implements differentiable `phase_imprecision` quantization with a straight-through estimator
- Implements Torch RNG based `phase_error` sampling with reproducible `torch.manual_seed(...)` behavior
- Averages probabilities, not amplitudes, for stochastic phase-error simulations
- Supports phase error with tensor superpositions, source noise, g2 `SectoredDistribution`, and post-measurement photon-loss transforms (verified with dedicated tests).
- Adds active-noise predicates and tightens noisy measurement validation.

Documentation & Benchmark:
- Ensures NoisySLOS can be executed on GPU
- Extends the noisy SLOS benchmark with forward-time reporting and optional noiseless baseline comparison

Tested for phase error and imprecision alone and tested with a mixture of all noise

> Merlin follows Perceval's forward phase-noise model. 
`phase_imprecision` quantizes each phase shifter to the nearest grid point with `round(phi / phase_imprecision) * phase_imprecision`. 
`phase_error` is then sampled as an additive uniform perturbation in `[-phase_error, phase_error]`.
When both are active, the sampled phase is `round(phi / phase_imprecision) * phase_imprecision + epsilon`. 

> So, forward phase computation is the same as Perceval, but the implementation differs from Perceval in execution model: Perceval applies phase noise by copying the circuit, writing quantized scalar phase values into `PS` components, and sampling `max_error * random.uniform(-1, 1)` when the unitary is computed. With Merlin, we apply the same forward computation with Torch tensors during `CircuitConverter.to_tensor()`, use Torch RNG, we supports batched parameters, and use a straight-through estimator so gradients continue to flow through the commanded phase. Moreover, for active `phase_error`, Merlin averages output probabilities over `n_phase_error_samples` sampled unitaries rather than averaging amplitudes.

> This is covered in tests 
- phase_imprecision matches Perceval exactly in tests/pcvl_pytorch/test_perceval_comparison.py::test_phase_imprecision_unitary_matches_perceval_oracle.
- Merlin’s own phase_error behavior is covered in tests/pcvl_pytorch/test_phase_noise_converter.py: Torch RNG reproducibility, different seeds, inactive flag, batched sampling, dtype/device, gradient flow, and combined phase_imprecision + phase_error
- Not for phase_error ecause Perceval uses Python random.uniform whereas MerLin use Torch RNG

## How to test / How to run

1. Command lines

```
pytest tests/
```

## Documentation
- [x] Docstrings updated
- [ ] Updated the API

## Checklist
- [x] Lint passes (ruff)


CassNot:
@codex review scalability and maintainability

merlin/algorithms/layer_utils.py
```diff
@@ -1069,42 +1131,53 @@ def validate_noisy_measurement_strategy(
...
+        )
+    if active_noise and output != "probabilities":
         raise ValueError(
             "When doing a noisy simulation, the probabilities measurement strategy must be used."
         )
-    if output == "mode_expectations" and noise is not None:
+    if not active_noise and output == "amplitudes":
         raise ValueError(
             "When doing a noisy simulation, the probabilities measurement strategy must be used."
         )
```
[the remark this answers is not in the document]
CassNot: True, it is completely misleading and it is edited in next commit

merlin/core/process.py
```diff
@@ -179,115 +248,379 @@ def _setup_computation_graphs(self):
...
+        active, this computes coherent output amplitudes for the sampled unitary,
+        converts those amplitudes to probabilities independently, and returns
+        those probabilities. Phase-error Monte Carlo averaging therefore uses
+        ``mean_k |SLOS(U_k) @ psi|^2`` rather than
+        ``|mean_k SLOS(U_k) @ psi|^2``.
+        """
+        if self._returns_probabilities():
+            return self._compute_source_probabilities_for_unitary(
+                unitary, amplitude_encoding=amplitude_encoding
+            )
```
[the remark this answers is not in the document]
CassNot: This is answering copilot comment above as well

- changed `process.py`: `_compute_probabilities_for_unitary()` routes to the incoherent source-noise mixture when `_has_source_noise()` is true -> therefore, `phase_error` without source noise leads to the coherent superposition path.
- updated `test_phase_noise_process.py` with explicit coherent and incoherent manual phase-error averages -> the tensor superposition regression now checks both that output matches coherent averaging + that it differs from the old `|c_i|^2` mixture.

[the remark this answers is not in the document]
CassNot: @codex review my last commit

merlin/algorithms/layer.py
```diff
@@ -191,9 +194,24 @@ def __init__(
             - ``MeasurementKind.PROBABILITIES`` returns a ``ProbabilityDistribution``
             - ``MeasurementKind.PARTIAL`` returns a ``PartialMeasurement``.
             - ``MeasurementKind.MODE_EXPECTATIONS`` returns a ``torch.Tensor``.
-        noise: pcvl.NoiseModel | None
-            The noise model used in the simulation. Default is None where no `noise` is
-            applied.
+        noise : pcvl.NoiseModel | None
+            Noise model used in the simulation. If omitted, no noise is
```
LF-Vigneux, line 201: I am not sure that a full description of a noise model implementation is the place here in the QauntumLayer's docstring. The other tests are described in detailed in the noisy simulations doc files. Maybe a link to the documentation. Otherwise I dont think a full description of the error is be be put here

merlin/core/process.py
```diff
@@ -179,115 +248,379 @@ def _setup_computation_graphs(self):
             self.simulation_graph, NoisySLOSComputeGraph
         ) or isinstance(self.simulation_graph, NoisyG2SLOSComputeGraph)
 
-    def compute(
+    def _default_fock_input_state(self) -> list[int]:
+        """Return the default fixed Fock state used for tensor input placeholders."""
+        return [1] * self.n_photons + [0] * (self.m - self.n_photons)
```
LF-Vigneux: If n_photons> n_modes ...

merlin/core/process.py
```diff
@@ -179,115 +248,379 @@ def _setup_computation_graphs(self):
...
+        sectors: list[SectorResult] = []
+        for left_sector in left.sectors:
+            right_sector = right.get_sector(left_sector.n_photons)
+            if left_sector.tensor.shape != right_sector.tensor.shape:
+                raise ValueError(
+                    "Phase-error sector tensor shape mismatch while averaging SectoredDistribution samples."
+                )
+            sectors.append(
+                SectorResult(
+                    left_sector.tensor + right_sector.tensor,
```
LF-Vigneux: keys may not match here, The object does not enforce a specific order...

merlin/core/process.py
```diff
@@ -179,115 +248,379 @@ def _setup_computation_graphs(self):
...
+
+            return mixed_probs
+
+        result = self.simulation_graph.compute_probs(unitary, self.input_state)
+        if isinstance(result, SectoredDistribution):
+            return result
+        _keys, probs = result
+        return probs
+
+    def _compute_superposition_amplitudes_for_unitary(
```
LF-Vigneux, line 398: Docstrings here between the compute functions you added/ rename would help a reviewer

merlin/pcvl_pytorch/locirc_to_tensor.py
```diff
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2025 Quandela
+# Copyright (c) 2026 Quandela
```
LF-Vigneux, line 3: We would need to change it in every file and add the MIT liscence to new files like noisy_slos.py

tests/algorithms/test_noise_contract.py
```diff
@@ -1,17 +1,21 @@
-import re
+from copy import deepcopy
```
LF-Vigneux, line 1: These tests in the file with return sectored distribution wont work once the g2 fix is merged, consider tranfroming them in a comment version to use tensor or wait for the fix to be merged

tests/core/test_phase_noise_process.py
```diff
@@ -0,0 +1,339 @@
...
+        )
+        accumulated = (
+            probabilities if accumulated is None else accumulated + probabilities
+        )
+
+    assert accumulated is not None
+    return accumulated / process._n_phase_error_samples
+
+
+def _manual_incoherent_phase_error_average(
```
LF-Vigneux, line 141: I dont undertsand the difference bewteen coherent and incoherent, both calculate the probs at the end. The end probs should indeed be the just the amplitudes square of the final state. I am not sure why they give different outputs

tests/pcvl_pytorch/test_phase_noise_converter.py
```diff
@@ -0,0 +1,450 @@
...
+
+    # Create a trainable phase parameter
+    phase = torch.tensor([0.5], dtype=torch.float64, requires_grad=True)
+
+    # Compute unitary with phase_error applied
+    torch.manual_seed(42)
+    unitary1 = converter.to_tensor(phase, apply_phase_error=True)
+
+    # Compute loss and backward pass
+    loss1 = unitary1.real.sum()
```
LF-Vigneux: Watch out with this sum for a loss

LF-Vigneux (changes requested):
Some changes requested to comments and test structures.

merlin/core/process.py
CassNot: There is indeed a silent behaviour here and if n_photons>n_modes then we would just have a list of [1] of size self.n_photons which is a bit problematic

CassNot: I am implementing a bugfix in this PR.
Now, when a default input state would need to be inferred and `n_photons > n_modes`, Merlin raises:
`Cannot derive a default input_state when n_photons is greater than n_modes. Provide an explicit input_state.
`
And I added tests for it

merlin/pcvl_pytorch/locirc_to_tensor.py
CassNot: I think we should start editing the files we edit with 2026 (I will verify that fact but I don't think it is blocking the rest)

merlin/core/process.py
```diff
@@ -513,16 +792,30 @@ def compute_with_keys(self, parameters: list[torch.Tensor]):
...
         -------
         tuple[Any, torch.Tensor]
-            Simulation-graph keys and corresponding probabilities if it is a noisy simulation and amplitude otherwise.
+            Simulation-graph keys and corresponding probabilities if source
+            noise or phase error is active, and amplitudes otherwise.
         """
+        if self._has_phase_error():
+            probabilities = self._compute_phase_error_probabilities(
+                parameters,
+                amplitude_encoding=isinstance(self.input_state, torch.Tensor),
```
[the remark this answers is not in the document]
CassNot: True but tensor `input_state` as well as `amplitude_encoding` input are deprecated in 0.4
Therefore, `compute_with_keys` should not decide based on tensor type and the intent should come from the layer/forward path that knows whether the input was a `StateVector` (and decides between angle and amplitude)
So to fix this bug, I am mirroring `compute`, but I don’t add new public `amplitude_encoding` (because of deprecation)

merlin/algorithms/layer.py
CassNot: I removed it :(

merlin/core/process.py
CassNot: True
This is unsafe because `right.get_sector(left_sector.n_photons)` only matches by photon count and does not guarantee that `right_sector.tensor[..., i]` corresponds to `left_sector.keys[i]`
And, the same risk exists in `_compute_source_probabilities_for_unitary`, where it does: `sector.tensor = sector.tensor + probs.get_sector(sector.n_photons).tensor`
I am fixing both

tests/core/test_phase_noise_process.py
CassNot: - A coherent error changes the quantum state while preserving phase relationships, so amplitudes can still interfere (unitary error)
- An incoherent error destroys or ignores phase relationships, so alternatives are combined as classical probabilities, not amplitudes (mixture)

I added some documentation support

merlin/core/state.py
```diff
@@ -42,6 +44,34 @@ class StatePattern(str, Enum):
...
+        raise ValueError(f"n_modes must be positive, got {n_modes}")
+
+    if n_photons < 0:
+        raise ValueError(f"n_photons must be non-negative, got {n_photons}")
+
+    if n_photons > n_modes:
+        raise ValueError(
+            "Cannot derive a default input_state when n_photons is greater than "
+            "n_modes. Provide an explicit input_state."
+        )
```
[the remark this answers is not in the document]
CassNot: No, this is a non trivial input_state, it could be [2,1,1,1] and not [5,0,0,0] for instance
As this is quite arbitrary, it should fail
amplitude_encoding will be removed and this will not fail

merlin/algorithms/layer_utils.py
```diff
@@ -1069,42 +1135,49 @@ def validate_noisy_measurement_strategy(
...
+    ):
         raise ValueError(
-            "When doing a noisy simulation, the probabilities measurement strategy must be used."
+            "Partial measurement is not supported with active noise. Use full measurement or disable noise."
         )
-    if output == "mode_expectations" and noise is not None:
+    if active_noise and output != "probabilities":
         raise ValueError(
             "When doing a noisy simulation, the probabilities measurement strategy must be used."
         )
```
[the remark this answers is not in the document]
CassNot: No: `phase_imprecision` is deterministic, but it is still classified as active circuit noise. The public QuantumLayer contract is that active noise requires `MeasurementStrategy.probs()`, independent of whether the lower-level process could technically still produce amplitudes.

CassNot (commented):
@claude review

CassNot:
@claude review

CassNot:
Answering @claude review

### For the first point addressing the `max_error` on PS component: fixed
Now, when a circuit contains at least `pcvl.PS(..., max_error>0)`:
- Merlin treats it as phase noise (even if no `NoiseModel.phase_error`)
- The PS is not precomputed as a constant tensor
- `to_tensor(..., apply_phase_error=True)` draws a fresh perturbation
- if a NoiseModel.phase_error is given, the max_error of the PS wins (with a Warning so users now [tested]) - as in Perceval
- tested with 2 PS with different max_error as well

### For the 2nd point addressing the MC simulation
It is true. To match Perceval contract, the nb of samples is by default 1
I documented the fact that noise is multiplicative across noise types

### For the 3rd point on keys
I added an explicit invariant check instead of threading keys through the whole phase-error averaging path
If they diverge, it raises: Probability keys returned by the simulation graph do not match the mapped output-key order.

merlin/core/process.py
```diff
@@ -179,115 +255,604 @@ def _setup_computation_graphs(self):
...
+            probs_stacked = torch.stack(tensor_probs_per_state, dim=0)
+            selected_weights = weights[:, active_indices].to(probs_stacked.dtype)
+            mixed_probs = torch.einsum("is,sbo->ibo", selected_weights, probs_stacked)
+
+            if mixed_probs.shape[1] == 1:
+                mixed_probs = mixed_probs.squeeze(1)
+
+            return mixed_probs
+
+        result = self.simulation_graph.compute_probs(unitary, self.input_state)
```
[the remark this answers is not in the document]
CassNot: OK because input torch tensors will be deprecated very soon so this case will not apply

docs/source/quantum_expert_area/noisy_simulations.rst
```diff
@@ -69,17 +69,170 @@ You can either add this noise model to a :class:`pcvl.Experiment` that is then u
...
+
+Merlin does not use this second expression for ``phase_error``.
+
+An incoherent error is represented as a classical mixture of alternatives. The
+relative phases between alternatives are not used for interference; Merlin
+combines probabilities, not amplitudes. Source noise is handled this way. For a
+tensor input state interpreted as a superposition, source-noise simulations
+propagate each active input basis state independently and combine the resulting
+probability distributions with weights :math:`|c_i|^2`.
+
```
LF-Vigneux, line 162: You could add an example: 

Averaging the amplitudes: $\frac{1}{\sqrt{2}}(\ket{10}+\ket{01})$ and $\frac{1}{\sqrt{2}}(\ket{10}-\ket{01})$, If you average the amplitudes, you only get $\ket{10}$ so the prbabilities is 1 at that state. If you average the probs you will get a 50/50 chance between the states.

merlin/core/process.py
```diff
@@ -430,102 +999,82 @@ def compute_ebs_simultaneously(
...
+        tuple[Any, torch.Tensor | SectoredDistribution]
+            Simulation-graph keys and corresponding probabilities if source
+            noise or phase error is active, and amplitudes otherwise.
         """
+        if self._has_phase_error():
+            probabilities = self._compute_phase_error_probabilities(
+                parameters,
+                amplitude_encoding=use_input_state_superposition,
+            )
+            if isinstance(probabilities, SectoredDistribution):
```
LF-Vigneux, line 1046: Here normally the simulation graph has all the keys no matter the noise

tests/pcvl_pytorch/test_phase_noise_converter.py
```diff
@@ -0,0 +1,535 @@
...
+
+    # Create a trainable phase parameter
+    phase = torch.tensor([0.5], dtype=torch.float64, requires_grad=True)
+
+    # Compute unitary with phase_error applied
+    torch.manual_seed(42)
+    unitary1 = converter.to_tensor(phase, apply_phase_error=True)
+
+    # Compute loss and backward pass (safe because 1x1 PS)
+    loss1 = unitary1.real.sum()
```
LF-Vigneux: It is not changed yet... The sum is not safe, use mse with squared amplitudes

LF-Vigneux (changes requested):
Couple of last changes

docs/source/quantum_expert_area/noisy_simulations.rst
CassNot: Added

merlin/core/process.py
CassNot: Not sure I understood correctly the comment but here is what I did:
- return keys and results when probs or results is a SD
- tested with coverage for source g2 and phase_error + g2

tests/pcvl_pytorch/test_phase_noise_converter.py
CassNot: for a 1x1 PS, an MSE on squared amplitudes is not safe if you expect a nonzero phase gradient, because` |exp(i phi)|^2 == 1` 
But in any case, I think it is better to test probs than amplitudes because we want to insure probabilities behaviour with noise
I am now using BS-PS-BS circuit for this case

LF-Vigneux (approved):
All good on my part

merlin/core/process.py
LF-Vigneux: No what I meant is that no nmatter the simulation graph, you have the keys as an attribute, but this is fine for now

PR #237, opened 2026-06-10
PML-351: MeasurementStrategy deprecations

LF-Vigneux:
## Summary
Complete the deprecation of the MeaurementStrategy enum object. Warnings are not errors.

## Related Issue
PML-351

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes
- Removed the enum type object for MeaurementStrategy.
   - Like it is mentionned in the ticket: Kept the MeasurmentStrategy.None object.
   - Respected in general the guidelines of deprecations at the top of the merlin/measurement/strategies.py file. Diverted from this plan to keep the None strategy
- Previous warnings that were thrown for enum-type measurement stratgies now fail and now public acess to the enum style probs, amplitude and mode expectation is now possible. 
- Cleaned the documentation so that no reference to enum-style measurement strategy is present.

## How to test / How to run
```
pytest -q
```

## Documentation
- [x] Docstrings updated
- [ ] Updated the API

### DIFF
--- a/merlin/measurement/strategies.py
+++ b/merlin/measurement/strategies.py
@@ -54,7 +54,7 @@ from merlin.utils.grouping import LexGrouping, ModGrouping
 # - If external compatibility is still needed, provide a separate shim module.
 
 
-### Note, kept some Legacy to keep the None measurement strategy
+# Note: kept some Legacy to keep the None measurement strategy
 
 
 class _LegacyMeasurementStrategy(Enum):
--- a/merlin/measurement/strategies.py
+++ b/merlin/measurement/strategies.py
@@ -35,7 +35,7 @@ import torch
 from merlin.core.computation_space import ComputationSpace
 from merlin.core.partial_measurement import PartialMeasurement
 from merlin.measurement.process import partial_measurement
-from merlin.utils.deprecations import warn_deprecated_enum_access
+from merlin.utils.deprecations import error_deprecated_enum_access
 from merlin.utils.grouping import LexGrouping, ModGrouping
 
 # Deprecation guide (target: v0.4):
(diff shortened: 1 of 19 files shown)


LF-Vigneux:
@claude

CassNot:
@claude review fixes

merlin/measurement/strategies.py
```diff
@@ -54,13 +54,13 @@
 # - If external compatibility is still needed, provide a separate shim module.
 
 
+# Note: kept some Legacy to keep the None measurement strategy
+
+
 class _LegacyMeasurementStrategy(Enum):
```
ben9871, line 60: this ticket is good but the computation space in quantumlayer also needs to be deprecated. leaving the comment here as this is a clarification of scope moreso than a review of the code already submitted

ben9871 (changes requested):

LF-Vigneux:
@claude review my new changes, now deprecated the argument in the quantumlayer's constructor

LF-Vigneux:
@ben9871 What do you think about this Claude comment:

Explicit None edge case: previously QuantumLayer(..., computation_space=None) was a silent no-op (it was the default); now passing it explicitly raises AttributeError because the key is present in kwargs. Acceptable for a removed arg, just noting the behavioral change for anyone who passed None explicitly.

Right now computation_space=None fails. I think it is alright since we remove the argument. What do you think?

ben9871:
> @ben9871 What do you think about this Claude comment:
> 
> Explicit None edge case: previously QuantumLayer(..., computation_space=None) was a silent no-op (it was the default); now passing it explicitly raises AttributeError because the key is present in kwargs. Acceptable for a removed arg, just noting the behavioral change for anyone who passed None explicitly.
> 
> Right now computation_space=None fails. I think it is alright since we remove the argument. What do you think?

seems good to me

ben9871 (approved):
changes done as requested, merged

PR #251, opened 2026-06-15
PML-352 Remove no_bunching flag

ben9871:
## Summary
Remove the deprecated `no_bunching` compatibility path for 0.4 and require explicit `ComputationSpace` configuration instead.

## Related Issue
Related Jira: PML-352

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [x] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling
- [x] Breaking change (requires migration notes)

## Proposed changes
- Hard-reject `no_bunching` for `QuantumLayer`, `QuantumLayer.simple`, `FidelityKernel`, `FidelityKernel.simple`, and `KernelCircuitBuilder.build_fidelity_kernel`.
- Replace the warning-backed `raise_no_bunching_deprecated` path with a no-warning `raise_no_bunching_removed` error.
- Remove `ComputationSpace.default(no_bunching=...)` and update internal test helpers to use explicit `ComputationSpace`.
- Update removal tests so `no_bunching` raises clearly without emitting `DeprecationWarning`.
- Update user/API docs and one reproduced-paper snippet to show explicit `ComputationSpace.UNBUNCHED` and `ComputationSpace.FOCK` usage.

## How to test / How to run
1. Focused removal and affected-path tests

```powershell
.\.venv\Scripts\python.exe -m pytest -p no:cacheprovider tests\core\test_unbunched.py tests\algorithms\test_kernels_deprecation.py tests\bridge\test_quantum_bridge.py tests\measurement\test_measurement_deprecations.py tests\measurement\test_measurement_strategy.py
```

2. Full tracked test suite

```powershell
$tests = git ls-files -- 'tests/**/test_*.py'
.\.venv\Scripts\python.exe -m pytest -p no:cacheprovider $tests
```

3. Pre-commit

```powershell
.\.venv\Scripts\python.exe -m pre_commit run --all-files
```

4. Docs

```powershell
$env:SPHINXOPTS='-W --keep-going -n'
$env:SPHINXBUILD='..\.venv\Scripts\sphinx-build.exe'
cd docs
.\make.bat clean
.\make.bat html
```

Focused tests: `104 passed, 4 skipped`.

Full tracked test suite: `961 passed, 106 skipped`.

Pre-commit: ruff, ruff format, and mypy passed.

Docs: Sphinx clean/html passed with `-W --keep-going -n`.

No expected runtime performance impact. The change removes legacy argument handling and updates tests/docs.

## Documentation
- [x] Docstrings updated
- [x] Updated the API

## Checklist
- [x] Lint passes (ruff)
- [x] With this command:

### DIFF
--- a/docs/source/api_reference/api/merlin.algorithms.kernels.rst
+++ b/docs/source/api_reference/api/merlin.algorithms.kernels.rst
@@ -29,10 +29,11 @@ Deprecations
    :undoc-members:
    :show-inheritance:
 
-.. warning:: *Deprecated since version 0.3:*
+.. warning:: *Removed in version 0.4:*
    The ``no_bunching`` flag accepted by legacy kernel constructors is removed
-   since version 0.3.0. Use the ``computation_space`` parameter instead.
-   See :doc:`/user_guide/migration_guide`.
+   in version 0.4. Use ``computation_space=ComputationSpace.UNBUNCHED`` or
+   ``computation_space=ComputationSpace.FOCK`` instead. See
+   :doc:`/user_guide/migration_guide`.
 
 .. warning:: *Deprecated since version 0.4:*
    Direct unitary construction through :meth:`FeatureMap.compute_unitary` is a
--- a/docs/source/api_reference/api/merlin.algorithms.layer.rst
+++ b/docs/source/api_reference/api/merlin.algorithms.layer.rst
@@ -414,9 +414,11 @@ For manual truncated backpropagation through time (TBPTT), set ``detach_at_each_
 
 Deprecations
 -------------------------
-.. warning:: *Deprecated since version 0.3:*
-   The use of the ``no_bunching`` flag  is deprecated and is removed since version 0.3.0.
-   Use the ``computation_space`` flag inside ``measurement_strategy`` instead. See :doc:`/user_guide/migration_guide`.
+.. warning:: *Removed in version 0.4:*
+   The ``no_bunching`` flag is removed in version 0.4. Use
+   ``MeasurementStrategy.probs(computation_space=ComputationSpace.UNBUNCHED)``
+   or ``MeasurementStrategy.probs(computation_space=ComputationSpace.FOCK)``
+   instead. See :doc:`/user_guide/migration_guide`.
 
 .. warning::
    *Deprecated since version 0.4:* The use of the ``computation_space`` argument in the QuantumLayer's constructor is no longer supported as 0.4.0.
--- a/docs/source/api_reference/api/merlin.utils.deprecations.rst
+++ b/docs/source/api_reference/api/merlin.utils.deprecations.rst
@@ -8,8 +8,8 @@ merlin.utils.deprecations module
 
 .. autofunction:: sanitize_parameters
 
-.. autofunction:: warn_deprecated_enum_access
+.. autofunction:: error_deprecated_enum_access
 
-.. autofunction:: raise_no_bunching_deprecated
+.. autofunction:: raise_no_bunching_removed
 
 .. autofunction:: normalize_measurement_strategy
--- a/docs/source/reproduced_papers/reproductions/photonic_memristor.rst
+++ b/docs/source/reproduced_papers/reproductions/photonic_memristor.rst
@@ -56,8 +56,9 @@ Memristor ``QuantumLayer`` usage (from the implementation):
        trainable_parameters=["theta"],
        input_parameters=["px"],
        input_state=[0, 1, 0],
-       measurement_strategy=ml.MeasurementStrategy.probs(),
-       no_bunching=True,
+       measurement_strategy=ml.MeasurementStrategy.probs(
+           computation_space=ml.ComputationSpace.UNBUNCHED
+       ),
    )
 
    phi_enc = encode_phase(x)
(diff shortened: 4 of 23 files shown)


ben9871:
@claude Review

ben9871:
> **Claude finished @ben9871's task in 1m 42s** —— [View job](https://github.com/merlinquantum/merlin/actions/runs/27535370552)
> 
> ### PR Review: Remove `no_bunching` flag
> * [x]  Read core changes (`deprecations.py`, `computation_space.py`, `process.py`, `slos_torchscript.py`, `feed_forward_legacy.py`)
> * [x]  Review test changes
> * [x]  Review docs changes
> * [x]  Post review feedback
> 
> Overall this is a clean, well-scoped removal. The hard-rejection is applied consistently across every entry point, the registry entries are uniform, the tests now correctly assert that **no** `DeprecationWarning` is emitted (not just that a `ValueError` is raised), and the docs are corrected. Nice work. A few small suggestions, none blocking.
> 
> ### 👍 Strengths
> * **Consistent rejection mechanism.** All `no_bunching` registry entries collapse to the same `(_NO_BUNCHING_REMOVED_MESSAGE, True, None)` shape, matching the existing fatal-deprecation convention (e.g. `ansatz`, `reservoir_mode`). Direct-call sites (`process.py`, `slos_torchscript.py`, `feed_forward_legacy.py`) all funnel through `raise_no_bunching_removed()`.
> * **Tests genuinely verify the new behavior.** Switching from `pytest.warns(DeprecationWarning)` to `warnings.catch_warnings(record=True)` + asserting no `DeprecationWarning` is the right call — a plain `pytest.raises(ValueError)` would have silently passed even if a stray warning leaked. Parametrizing over `[True, False]` is a good addition.
> * **Docs fix a real inaccuracy.** The old text claimed `no_bunching` was _"removed since version 0.3.0"_, which was wrong (it was only deprecated). The new "Removed in version 0.4" wording is correct and consistent across `migration_guide.rst`, `computation_space.rst`, and the docstrings.
> * **`feed_forward_legacy.py` simplification is correct.** After `raise_no_bunching_removed()`, `no_bunching` can only be `None`, whose old default was `True → UNBUNCHED`, so hardcoding `ComputationSpace.UNBUNCHED` preserves behavior. Removing the dead `coerce(... if no_bunching else ...)` branch is the right cleanup.
> 
> ### 🔧 Suggestions (non-blocking)
> **1. `raise_no_bunching_removed()` should be typed `NoReturn`** — `merlin/utils/deprecations.py:38` The function always raises and the docstring even states _"This function never returns normally"_, but it's annotated `-> None`. Typing it `from typing import NoReturn` / `-> NoReturn` lets mypy understand that code after the call is unreachable (e.g. it would let you drop the now-redundant fall-through logic elsewhere), and documents intent precisely.
> 
> ```python
> from typing import NoReturn
> 
> def raise_no_bunching_removed() -> NoReturn:
>     raise ValueError(_NO_BUNCHING_REMOVED_MESSAGE)
> ```
> 
> [Fix this →](https://claude.ai/code?q=On%20branch%20PML-352%2C%20in%20merlin%2Futils%2Fdeprecations.py%2C%20change%20raise_no_bunching_removed%20return%20type%20annotation%20from%20None%20to%20NoReturn%20%28import%20NoReturn%20from%20typing%29%20since%20it%20always%20raises%2C%20and%20update%20its%20docstring%20Returns%20section%20accordingly&repo=merlinquantum/merlin)
> 
> **2. String-equality special-casing is a mild code smell** — `merlin/utils/deprecations.py:319-322`
> 
> ```python
> if msg == _NO_BUNCHING_REMOVED_MESSAGE:
>     base = msg
> else:
>     base = f"Parameter '{key}' is deprecated. {msg}"
> ```
> 
> Comparing against a specific message string to decide formatting won't scale once a second "removed" (as opposed to "deprecated") parameter appears — and the word "deprecated" in the prefix would be semantically wrong for a _removed_ param anyway. Consider encoding this in the registry itself rather than by message identity, e.g. a distinct severity sentinel (a `"removed"` value alongside `True`/`False`/`None`) or a 4th "pre-formatted" flag in the tuple. Fine to defer, but worth a `# TODO` if not addressed now.
> 
> ### ℹ️ Out-of-scope observations (pre-existing, not introduced here)
> * **`normalize_measurement_strategy` has an unreachable-but-latent bug** — `deprecations.py:410` references `computation_space`, which is never bound on the fall-through path (the function has no such parameter and only the early branches assign it). If an input ever slips past the `None`/`str`/`MeasurementStrategy`/`MeasurementKind` checks, this raises `UnboundLocalError` rather than a clear error. Not touched by this PR, but it lives in the same file — worth a follow-up.
> * **`modern_factory_raises_no_error`** in `tests/measurement/test_measurement_deprecations.py:219` lacks a `test_` prefix (and isn't a `@staticmethod`/`self` method), so pytest never collects it. This PR only stripped unused `layer =` assignments inside it; the dead-test issue predates this change but means those assertions aren't actually running.
> 
> ### Verification
> I reviewed the diff statically and did not run the suite here, but the test changes look correct and the PR reports `104 passed, 4 skipped` (focused) and `961 passed, 106 skipped` (full) with clean pre-commit and `-W` docs build. · `PML-352`

I think the first change here is justifiable, however, I think the rest are out of scope

LF-Vigneux (approved):
Everything seems ok to me!

PR #283, opened 2026-07-13
PML-380: Unitary device fix for g2

LF-Vigneux:
## Summary
Fixed this issue:

Noisy g2 path allocates on self.device/self.dtype, not unitary.device — merlin/pcvl_pytorch/noisy_slos.py:238-258,316. p_emit/weight_k and each SectorResult.tensor are built on self.device, while the sub-graph probs live on unitary.device. A CUDA unitary on a graph constructed with default device=None (without a prior .to()) raises a cross-device error. The non-g2 path correctly derives device from unitary.device (:768) — mirror that here.

## Related Issue
PML-380

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes
- Initialized the g2 and sectors results with the unitary's device instead of the SLOS'.

## How to test / How to run
```
pytest -q
```

## Documentation
- [ ] Docstrings updated
- [ ] Updated the API

### DIFF
--- a/tests/pcvl_pytorch/test_noisy_slos.py
+++ b/tests/pcvl_pytorch/test_noisy_slos.py
@@ -372,7 +372,7 @@ def test_noisy_g2_slos_unitary_cuda_graph_cpu():
     )
 
     unitary = torch.tensor(
-        np.array(pcvl.Matrix.random_unitary(m=5), dtype=np.complex128),
+        np.array(pcvl.Matrix.random_unitary(n=5), dtype=np.complex64),
         device=torch.device("cuda"),
     )
 
--- a/tests/pcvl_pytorch/test_noisy_slos.py
+++ b/tests/pcvl_pytorch/test_noisy_slos.py
@@ -7,6 +7,7 @@ from merlin import Combinadics, ComputationSpace
 from merlin.algorithms.layer_utils import NoiseGroups, classify_noise
 from merlin.pcvl_pytorch.locirc_to_tensor import CircuitConverter
 from merlin.pcvl_pytorch.noisy_slos import (
+    NoisyG2SLOSComputeGraph,
     NoisySLOSComputeGraph,
     _InputStateNoisySLOSComputeGraph,
 )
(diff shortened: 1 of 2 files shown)


LF-Vigneux:
@claude review

tests/pcvl_pytorch/test_noisy_slos.py
```diff
@@ -360,3 +361,20 @@ def test_computation_space_and_indistinguishability_default_value(
...
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+def test_noisy_g2_slos_unitary_cuda_graph_cpu():
+    noise = pcvl.NoiseModel(indistinguishability=0.2, g2=0.2)
+    groups = classify_noise(noise)
+    noisy_slos = NoisyG2SLOSComputeGraph(
+        groups, m=5, n_photons=3, computation_space=ComputationSpace.FOCK, device=None
+    )
+
+    unitary = torch.tensor(
+        np.array(pcvl.Matrix.random_unitary(m=5), dtype=np.complex128),
```
CassNot: here it should be `n=5` and not `m=5` otherwise it fails with a perceval error

CassNot: Another error from this line is that the type is `np.complex128` which is inconsistent with the `torch.float32` defaults types from the compute graph. If you set it, you need to change the type of the graph. My fix would be to just have the `dtype=np.complex64` here and that works !

CassNot (changes requested):
The fix does not work (cuda issue) once the perceval typo and type issue are solved (`n=5` and `dtype=np.complex64`)

You will find my fix in the file attached:
- lines 214-215 and 225-226 and 232-238: I am ensuring the probs stay on the same device as the unitary (if g2_distinguishable or not)
- lines 783 and 796, I am also sending the bit_weight and convolution to the unitary.device

So now, all probability tensors are explicitly placed on the unitary's device before use. This should prevent device mismatch errors during convolution and other operations. 

With all these fixes, it works on GPU

[noisy_slos.py](https://github.com/user-attachments/files/29984719/noisy_slos.py)

CassNot:
@claude I believe your analysis was wrong and the code was not working on GPU because 
- when `compute_probs` is called with a unitary on a different device, there's a device mismatch
- when computing one-hot state probabilities, they're not being moved to the unitary's device before convolution
All probability tensors should be explicitly placed on the unitary's device before use, preventing device mismatch errors during convolution

CassNot (approved):
All tests now pass on GPU !

PR #284, opened 2026-07-14
PML-388: fixed random phase initialization in entangling layer if trainable is False

CassNot:
## Summary

Non-trainable entangling layers (`add_entangling_layer(trainable=False, ...)`) were silently pinning every inner/outer phase shifter to `0.0`, collapsing the block into a deterministic swap/identity instead of a random fixed unitary. This broke architectures (e.g. reservoir-computing style circuits) that rely on non-trainable entangling layers to inject a fixed random mix. `GenericInterferometer` now draws random fixed phases for any non-trainable phase shifters at construction time, optionally seeded for reproducibility.

This also fixes a combinatorial bug in `NoisyG2SLOSComputeGraph` . g2 extra-photon sectors were computed by growing a whole new SLOS graph per input state instead of reusing the base Orthogonal Bad Bits (OBB) partitions, causing incorrect probabilities/tests to rely on the degenerate zero-phase circuits. This bug was highlighted upon solving the first one (the tests were failing because the circuit was not a swap circuit anymore). This fix is inspired by Perceval management of g2 extra photon.

## Related Issue
   
PML-388

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / Cleanup
- [ ] Performance improvement
- [ ] CI / Build / Tooling

## Proposed changes

- `merlin/core/components.py`: `GenericInterferometer` gains `seed`, `fixed_inner_values`, and `fixed_outer_values` fields. `__post_init__` now draws random phases (via `random.Random(seed)`) for any non-trainable inner/outer phase shifters instead of hardcoding `0.0`.
- `merlin/builder/circuit_builder.py`: `add_entangling_layer(..., seed=None)` forwards the seed to `GenericInterferometer`; the generated Perceval circuit now uses `block.fixed_inner_values`/`fixed_outer_values` for non-trainable phase shifters instead of literal `0.0`.
- `merlin/pcvl_pytorch/noisy_slos.py`: Added `NoisyG2SLOSComputeGraph._augmented_obb_probs`, which reuses the cached OBB partitions of the base input state and grows only the coherent cell's Fock state by the g2-emitted photons, replacing the previous approach of running a full new SLOS graph per extra-photon combination. Fixes incorrect probabilities for g2 extra-photon sectors.
- Test suite updated to reflect that non-trainable layers now carry non-zero random phases, plus new regression tests for reproducibility (seeded vs. unseeded), partial trainable/non-trainable mixes, and the `NoisyG2SLOSComputeGraph` fix.

## How to test / How to run

1. Command lines

```
pytest tests/builder/test_circuit_builder.py tests/core/test_components.py tests/core/test_merlin_processor_unit.py -v
pytest tests/pcvl_pytorch/test_noisy_g2_slos.py -v
```

No expected performance regression; `_augmented_obb_probs` reuses cached SLOS subgraphs instead of allocating new ones per combination, so it should be at least as efficient as the previous implementation.

## Documentation
- [x] Docstrings updated
- [ ] Updated the API

## Checklist
- [x] Lint passes (ruff)

### DIFF
--- a/tests/builder/test_circuit_builder.py
+++ b/tests/builder/test_circuit_builder.py
@@ -399,6 +399,9 @@ def test_non_trainable_entangling_layer_uses_random_phases_not_zero():
     # Phases must be populated and not trivially all-zero.
     assert component.fixed_inner_values
     assert component.fixed_outer_values
+    # random.uniform() draws from a continuous distribution; all values being
+    # exactly 0.0 has measure-zero probability and is effectively impossible.
+    # These assertions guard against the old bug where phases were hardcoded to 0.0.
     assert any(v != 0.0 for v in component.fixed_inner_values)
     assert any(v != 0.0 for v in component.fixed_outer_values)
     for v in component.fixed_inner_values + component.fixed_outer_values:
@@ -457,6 +460,8 @@ def test_non_trainable_entangling_layer_partial_trainable_mix():
     assert component.trainable_outer is False
     assert component.fixed_inner_values == []
     assert component.fixed_outer_values
+    # random.uniform() draws from a continuous distribution; all values being
+    # exactly 0.0 has measure-zero probability and is effectively impossible.
     assert any(v != 0.0 for v in component.fixed_outer_values)
 
     pcvl_circuit = builder.to_pcvl_circuit(pcvl)
--- a/tests/builder/test_circuit_builder.py
+++ b/tests/builder/test_circuit_builder.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import math
 import os
 from pathlib import Path
 
+import numpy as np
 import perceval as pcvl
 import pytest
 import torch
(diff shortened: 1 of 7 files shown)


CassNot:
@claude review

CassNot:
@claude review

tests/builder/test_circuit_builder.py
```diff
@@ -378,6 +380,91 @@ def test_entangling_layer_mode_range_and_non_trainable():
...
+    builder.add_entangling_layer(
+        name="mix", trainable_inner=True, trainable_outer=False
+    )
+
+    component = builder.circuit.components[-1]
+    assert component.trainable_inner is True
+    assert component.trainable_outer is False
+    assert component.fixed_inner_values == []
+    assert component.fixed_outer_values
+    assert any(v != 0.0 for v in component.fixed_outer_values)
```
LF-Vigneux, line 465: Technically could be 0...

tests/builder/test_circuit_builder.py
```diff
@@ -378,6 +380,91 @@ def test_entangling_layer_mode_range_and_non_trainable():
...
+    builder.add_entangling_layer(trainable=False, name="pre_mix")
+
+    component = builder.circuit.components[-1]
+    assert isinstance(component, GenericInterferometer)
+    assert component.trainable is False
+
+    # Phases must be populated and not trivially all-zero.
+    assert component.fixed_inner_values
+    assert component.fixed_outer_values
+    assert any(v != 0.0 for v in component.fixed_inner_values)
```
LF-Vigneux, line 405: They technically can be 0 making this assertion false.

tests/core/test_components.py
```diff
@@ -34,3 +39,80 @@ def test_beam_splitter_get_params_exposes_non_fixed_names():
...
+def test_generic_interferometer_non_trainable_gets_random_phases():
+    """Non-trainable phase shifters must draw random phases, not default to 0.0."""
+    gi = GenericInterferometer(start_mode=0, span=4, trainable=False)
+
+    assert gi.trainable is False
+    count = gi.span * (gi.span - 1) // 2
+    assert len(gi.fixed_inner_values) == count
+    assert len(gi.fixed_outer_values) == count
+    assert any(v != 0.0 for v in gi.fixed_inner_values)
+    assert any(v != 0.0 for v in gi.fixed_outer_values)
```
LF-Vigneux, line 56: They can technically be 0 making this assrtion False.

LF-Vigneux (changes requested):
Couple of changes in tests, everything else seems sound

tests/builder/test_circuit_builder.py
CassNot: But then they would have to be all 0 and it is not possible that `random.uniform()` leads to all 0 because it draws from a continuous uniform distribution, so the probability of getting exactly 0.0 is measure-zero

tests/builder/test_circuit_builder.py
CassNot: Same explanation as above

tests/core/test_components.py
CassNot: Same explanation as above

CassNot:
@LF-Vigneux, I do not think it is possible for all values to be 0 because they are samples from a uniform distribution
I added comments in the test files you mentioned 

@claude review

LF-Vigneux (approved):
I approve

PR #292, opened 2026-07-30
Fix large code block in PR template

CassNot:
- [ ] With this command: 

is becoming
- [ ] With this command: `SPHINXOPTS="-W --keep-going -n" make -C docs clean html` the docs are built without any warning or errors.

This avoids the PR description being too large from the large space between all bullet points in the checklist


LF-Vigneux (approved):
I approve

